PR #3, opened 2018-07-20
Fix LICENSE

secondsun:
The License should be APL 2, not MIT.  The Copyright should be Red Hat Inc. Not Dara Hayes


secondsun:
@aidenkeating rebased.

PR #4, opened 2018-07-23
AEROGEAR-7680 Add template for mobile client service

aidenkeating:
Add a client for the mobile client service along with information on how to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,23 @@ make ui
 
 ## Run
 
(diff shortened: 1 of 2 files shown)


secondsun:
Looking

secondsun:
Well the template imported cleanly.  Not sure if there is anything else I'm supposed to check, but +1 on that and on the readme changes.

PR #10, opened 2018-07-25
Use echo middleware

aidenkeating:
@wei-lee @sedroche Would you mind taking a look? `echo` already provides SPA functionality in its static middleware. So I moved from `gorilla/mux` to this, echo has a larger community also.

The current way it works:

`/ui/*` -> Everything under here is the UI, the user needs to navigate to `/ui/mobileclient/2` to retrieve a mobile client.
`/hello` -> API routes are straight from the root.

The way it works now:

`/*` -> It'll look for a static file, if it receives a `404` on the file search then it will return the React SPA.
`/api/*` -> All backend handlers are grouped into an `/api/` group, if a `404` is encountered here it'll be passed back to the user. We can also create a subgroup of the `/api/` group to allow for versioning if we want (e.g. `/api/v1/`).

This also removes the need to have our own logging middleware and pass the logger around so much. Instead the logging middleware in `echo` is used and a logger is provided with the request context.

I had to `curl` the API endpoints as I think the browser does some weird caching. But it does seem to work.


wei-lee:
👀

aidenkeating:
Replaced by https://github.com/aerogear/mobile-client-service/pull/11

PR #12, opened 2018-07-25
Aerogear 7693 - Create Mobile Client Service Chart

ciaranRoche:
## Motivation 

JIRA -  https://issues.jboss.org/browse/AEROGEAR-7693

## Description
Based on top of @sedroche fork, this change adds a Mobile Service Chart Component. The component expects two props, the number of unbound service and bound services.

## Screenshot
mobile-services1
*NOTE* : gif does not capture the mouse pointer correctly
## Note
Might be best to wait on @sedroche to be merged first and rebase/refactor this PR as needs be, then.


ui/src/MobileClientServiceChart.js
```diff
@@ -0,0 +1,44 @@
+import React, { Component } from 'react';
+import { DonutChart } from 'patternfly-react';
+
+import './styles/MobileClientServiceChart.css'
+
+class MobileClientServiceChart extends Component {
+    viewAllServices(){
```
sedroche: This is not really part of this component. We can move it when we integrate all the components though

PR #13, opened 2018-07-25
Replace dep with glide

wei-lee:
# Why

The k8s go-client and openshift go-client don't support dep at the moment. To make it easier to use those libs, we want to switch to glide instead.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -7,7 +7,7 @@ import (
 )
 
 func TestConfig(t *testing.T) {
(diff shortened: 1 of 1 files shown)


wei-lee:
ping @aidenkeating mind taking a look?

wei-lee:
@maleck13 would appreciate if you can keep an eye on this

maleck13:
@wei-lee shouldn't need to use glide the operator stuff all uses dep as does service catalog

maleck13:
https://github.com/aerogear/keycloak-operator/blob/master/Gopkg.toml

PR #16, opened 2018-07-25
Aerogear 7703 - folder restructure and added sass

ciaranRoche:
## Motivation

JIRA - https://issues.jboss.org/browse/AEROGEAR-7703

## Description
Adds support to react app to allow the use of sass.

Change also includes a cleaner folder structure, adding a components folder. 

## Note
This change will mean heavy rebasing on any work that is currently in progress, open to suggestions on the order of when to merge this change to minimize conflicts.


ui/package.json
```diff
@@ -3,6 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "node-sass-chokidar": "^1.3.3",
```
camilamacedo86: IHMO we should not use `^` since it can bring break changes which may impact the project.

ui/src/components/App.js
```diff
@@ -12,7 +12,7 @@ class App extends Component {
         <div className="App">
           <Switch>
             <Route exact path="/overview" component={Overview}/>
-            <Route exact path="/mobileclient/:id" component={Client}/>
+            <Route exact path="/mobileclien/:id" component={Client}/>
```
sedroche: Typo

ui/src/components/App.scss
```diff
@@ -0,0 +1,2 @@
+@import "../style/variables";
```
sedroche: The default layout would have this in `index.scss`. Why is it here now?

PR #19, opened 2018-07-26
Create the mobile configuration view

jhellar:
# Motivation
https://issues.jboss.org/browse/AEROGEAR-7666

screenshot

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -8,6 +8,7 @@
     "patternfly": "^3.52.2",
     "patternfly-react": "^2.11.1",
     "react": "^16.4.1",
+    "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.4.1",
     "react-router-dom": "^4.3.1",
     "react-scripts": "1.1.4"
--- a/ui/package.json
+++ b/ui/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "classnames": "^2.2.6",
     "node-sass-chokidar": "~1.3.3",
     "npm-run-all": "^4.1.3",
     "patternfly": "^3.52.2",
--- a/ui/src/components/ConfigurationView.js
+++ b/ui/src/components/ConfigurationView.js
@@ -5,6 +5,8 @@ import ServiceSDKInfo from './ServiceSDKInfo';
 import InlineEdit from './InlineEdit';
 import CopyToClipboardMultiline from './CopyToClipboardMultiline';
 
+import './ConfigurationView.css';
+
 const config = `{
   "version": 1,
   "clusterName": "https://192.168.0.106:8443",
@@ -49,7 +51,7 @@ class ConfigurationView extends Component {
           </Col>
           <Col md={6}>
             <Header>Mobile Client Config</Header>
-            <CopyToClipboardMultiline style={{ marginBottom: 25 }}>
+            <CopyToClipboardMultiline className="mobile-client-config">
               {config}
             </CopyToClipboardMultiline>
           </Col>
@@ -59,8 +61,8 @@ class ConfigurationView extends Component {
             <Header>SDK Configuration</Header>
             <h4><a>Android SDK Setup</a></h4>
           </Col>
-          <Col md={12} style={{ padding: 0 }}>
-            <Header style={{ margin: 20 }}>Service Configuration</Header>
+          <Col md={12} className="service-configuration">
+            <Header className="service-configuration-header">Service Configuration</Header>
             {
               serviceSDKInfo.map((info, index) =>
                 <ServiceSDKInfo
--- a/ui/src/components/ConfigurationView.js
+++ b/ui/src/components/ConfigurationView.js
@@ -1,0 +1,83 @@
+import React, { Component } from 'react';
+import { Grid, Row, Col } from 'patternfly-react';
+import Header from './Header';
+import ServiceSDKInfo from './ServiceSDKInfo';
+import InlineEdit from './InlineEdit';
+import CopyToClipboardMultiline from './CopyToClipboardMultiline';
+
+const config = `{
+  "version": 1,
+  "clusterName": "https://192.168.0.106:8443",
+  "namespace": "myproject",
+  "clientId": "myapp-android",
+  "services": []
+}`;
+
+const serviceSDKInfo = [
+  {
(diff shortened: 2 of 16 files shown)


ui/src/components/ConfigurationView.js
```diff
@@ -0,0 +1,69 @@
...
+        </Row>
+        <Row>
+          <Col md={12}>
+            <Header>SDK Configuration</Header>
+            <h4><a>Android SDK Setup</a></h4>
+          </Col>
+          <Col md={12} style={{ padding: 0 }}>
+            <Header style={{ margin: 20 }}>Service Configuration</Header>
+            <ServiceSDKInfo
+              serviceLogoUrl="https://pbs.twimg.com/profile_images/702119821979344897/oAC05cEB_400x400.png"
```
ciaranRoche: I am not sure of the requirements of this ticker, but should we be passing an array of service info and maybe mapping over it to build out the service sdk info.

ui/package.json
```diff
@@ -8,6 +8,7 @@
     "patternfly": "^3.52.2",
     "patternfly-react": "^2.11.1",
     "react": "^16.4.1",
+    "react-copy-to-clipboard": "^5.0.1",
```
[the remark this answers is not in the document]
jhellar: This should allow patch and minor updates [1], which should not introduce any breaking change.

[1] https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

ui/src/components/ConfigurationView.js
```diff
@@ -0,0 +1,69 @@
...
+        <Row>
+          <Col md={6}>
+            <Header>Mobile Client Details</Header>
+            <dl className="dl-horizontal left">
+              <dt>Client Type:</dt>
+              <dd>android</dd>
+              <dt>Client ID:</dt>
+              <dd>com.aerogear.androidshowcase</dd>
+              <dt>Client API Key:</dt>
+              <dd>b331aa8c-6c39-5f07-81f8-9eccf7fd6702</dd>
```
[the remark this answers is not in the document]
jhellar: These values will be fetched from the backend. I think that there will be separate ticket for this, once it is implemented in backend.

jhellar:
@ciaranRoche thanks for review, should be fixed now.

ui/src/components/CopyToClipboardMultiline.js
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import { OverlayTrigger, Tooltip } from 'patternfly-react';
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+class CopyToClipboardMultiline extends Component {
+  render() {
+    return (
+      <div className="input-group" style={{ display: 'block', overflowX: 'auto', ...this.props.style }}>
```
sedroche: So far we are using classes to define our styling. We can use the patternfly spacing and colors that way.
Can you move the inline styling to scss?

ui/src/components/ConfigurationView.js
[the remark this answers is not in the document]
sedroche: This is a mock mobile client. You can match it to this.

```.json
			{
					"kind": "MobileClient",
					"metadata": {
							"annotations": {
									"aerogear.org/service-instance-name": "dh-android-app-apb-4hxhp",
									"icon": "fa fa-android"
							},
							"clusterName": "",
							"creationTimestamp": "2018-07-25T21:10:17Z",
							"name": "myapp-android",
							"namespace": "test",
							"resourceVersion": "2669",
							"selfLink": "/apis/mobile.k8s.io/v1alpha1/namespaces/test/mobileclients/myapp-android",
							"uid": "21830e50-904f-11e8-8ef3-c85b76719196"
					},
					"spec": {
							"apiKey": "874e600e-3527-5e2a-a0a4-a530f7899c74",
							"appIdentifier": "dwqwdwqdw",
							"clientType": "android",
							"name": "myapp"
					}
			}
```

ui/src/components/Header.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
```
sedroche: I have something similar coming in my branch => https://github.com/aerogear/mobile-client-service/blob/over-list-base/ui/src/components/common/ComponentSectionLabel.js / https://github.com/aerogear/mobile-client-service/blob/over-list-base/ui/src/components/common/ComponentSectionLabel.scss

(Naming could probbaly be improved) but it is what is defined in openshift currently

ui/src/style/index.scss
```diff
@@ -1,2 +1,6 @@
 @import "./variables";
 @import "patternfly";
+
+.dl-horizontal.left dt {
+  text-align: left;
```
sedroche: There is a bootstrap class `text-left` I think.

ui/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "classnames": "^2.2.6",
```
camilamacedo86: The `ˆ` should not be used since it can install a major version which can impact the project with break changes

jhellar: See my answer ^

ui/src/components/Header.js
jhellar: Ok, should I leave it like this for now and update it so that it uses your component later?

jhellar:
@sedroche thanks for review. I've updated the PR so that it uses scss instead of inline styles. Can you review again?

ui/package.json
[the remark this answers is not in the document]
camilamacedo86: @jhellar following some parts of the doc added by you. 
 
`The `ˆ` : Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch]`

`Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.`

Because of it IHMO the best approach is to use `~` instead of `^`, besides by concept the break changes according to SemVer be Just in the X.0.0 some authors may add it in 0.X.0.

ui/src/components/ConfigurationView.js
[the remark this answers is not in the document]
camilamacedo86: 👍

ui/src/components/Header.js
sedroche: :+1:

jhellar:
@sedroche should be fixed now.

PR #20, opened 2018-07-26
add mobile client notification component

ciaranRoche:
## Motivation 

JIRA -  https://issues.jboss.org/browse/AEROGEAR-7710

## Description
Add notification alert for un bound services.

## Screenshot
screenshot from 2018-07-26 16-42-20


jhellar:
Two comments, otherwise looks good to me (haven't tested it yet):

- the `render` method does not need to use class field syntax. I think we should decide whether we want to bind additional methods in constructor or use class field syntax. According to [1][2] class field syntax is experimental. 

- not sure about the `view` variable, should not it be implemented as the component's state?

[1] https://reactjs.org/docs/handling-events.html
[2] https://github.com/facebook/react/issues/9851

ciaranRoche:
@jhellar good catch, I didn't know the class field syntax is experimental, I addressed that in the latest commit.
As for the `view` variable, its a workaround as this notification will be under a patternfly `ListView` and the behavior wanted is, if the notification is dismissed it will remain dismissed regardless of the ListView being toggled, Moving the variable outside of the component seems to satisfy this. Open to suggestions?

sedroche:
@ciaranRoche Im not too sure on the details of how react rendering works but this looks like a module level variable to me. So if we have multiple `<MobileClientNotification></MobileClientNotification>`s in a view they will be sharing the state?

ui/src/components/MobileClientListViewItem.js
```diff
@@ -0,0 +1,167 @@
...
+import MobileClientServiceChart from './MobileClientServiceChart';
+import ComponentSectionLabel from './common/ComponentSectionLabel';
+import MobileClientBuildList from './MobileClientBuildList';
+
+import './OverviewListItemHeader.css';
+
+// todo 
+const mobileServices = {
+    "mobileServices": {
+        "bound": 1,
```
camilamacedo86: What mean this numbers 1 and 2? Could be they in a const with a name in order to make clear their propose?

ciaranRoche: Its just dummy data for now to populate the patternfly donut chart. This will be solved later on.

camilamacedo86: 👍

PR #21, opened 2018-07-26
[WIP] Add k8s go client, and implement APIs for listing service instances, builds and buildconfigs

wei-lee:
# What

This PR adds a few dependencies to allow interacting with the k8s, service catalog and openshift APIs.
At the moment, 3 new routes are added to return the data that will be required by the UI:

* `/api/serviceinstances`
* `/api/builds`
* `/api/buildconfigs`

At the moment, those endpoints doesn't do too much. They just proxy the response back from the OpenShift cluster. There is not auth/authz checks. It will just load the kube config file and use the current context to talk to the cluster. 

# Why 

To setup the basic APIs and the dependencies to allow the UI to work with real data. 

# Try it

If you want to try this, please do the following:

1. Setup your local oc cluster and login via `oc`
2. setup the `NAMESPACE` env var to point to a project you want to test with
3. run `make setup && make serve`
4. once the server is running, try curl the above endpoints, and you should see some response back.

# Tasks

[x] Implementation
[ ] Unit tests
[ ] Integration tests

@sedroche can you see if this can be used by the UI yet? 

@StephenCoady @secondsun @ciaranRoche @wojta @jhellar @pawelpaszki @ziccardi Feel free to take a look at this PR and give feedback. I am new to Golang as well so this implementation may not necessary be the best way to do it. But at least we have something in place that we can iterate on. Also feel free to add more tests if you want to. Here is an example of how to test the k8s client: https://github.com/kubernetes/client-go/blob/master/examples/fake-client/main_test.go.


sedroche:
@wei-lee I will try to test this out with the UI

PR #24, opened 2018-07-30
Over list base

sedroche:
overview

Adding the list of clients to the overview view cc @ciaranRoche @StephenCoady 

@jhellar Can you take a look?


pkg/web/router.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	"github.com/labstack/echo"
 	"github.com/labstack/echo/middleware"
+	"github.com/labstack/echo"
```
StephenCoady: I think this is breaking the build

pkg/web/router.go
```diff
@@ -5,6 +5,8 @@ import (
...
+	"strings"
```
StephenCoady: same here

PR #26, opened 2018-07-31
download component complete

pawelpaszki:


ui/src/components/overview/BuildDownloadComponent.scss
```diff
@@ -0,0 +1,15 @@
...
+ span, img {
+  font-weight: 600;
+  display: inline-block;
+  vertical-align: top;
+}
+ img {
+  margin-left: 5em;
+}
+ .build-download-component {
+  width: 100%;
```
sedroche: `div` should be naturally 100%

sedroche:
@pawelpaszki Can you add an image of the component please?

Is there a react component that creates the QR codes?

pawelpaszki:
@sedroche I have downloaded a placeholder for the QR Code image. I thought it is good for now. We should probably find out, how is the QR Code generated in OpenShift. Any thoughts?

sedroche:
@pawelpaszki The qr code was a custom component that we created => https://github.com/aerogear/origin-web-console/blob/aerogear-mcp/app/scripts/directives/jqQrCode.js based on https://github.com/aerogear/origin-web-console/blob/aerogear-mcp/bower.json#L23

If no react component exists we will have to create our own.

pawelpaszki:
@sedroche would you mind taking a look?

ui/src/components/overview/MobileClientBuildList.js
```diff
@@ -29,6 +30,10 @@ class MobileClientBuildList extends Component {
                                             created <Moment fromNow>{mobileClientBuild.status.startTimestamp}</Moment>
                                     </span>
                                 </Col>
+                                <Col md={12}>
+                                    <BuildDownloadComponent downloadURL={mobileClientBuild.downloadURL} QRSize={256}/>
+                                </Col>
+
```
StephenCoady: We can remove these lines and the import at the top altogether, I will integrate this component on my branch.

ui/src/components/overview/BuildDownloadComponent.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import { Alert, Col} from 'patternfly-react';
+import './BuildDownloadComponent.css';
+var QRCode = require('qrcode.react');
+
+class BuildDownloadComponent extends Component {
```
sedroche: What do you think of `<BuildDownloadLinks>` as a name?

ui/src/components/overview/BuildDownloadComponent.js
```diff
@@ -0,0 +1,30 @@
...
+
+    render() {
+        const size = this.props.QRSize || 256;
+        return (
+            <div className="build-download-component">
+                <Col md={12}>
```
sedroche: The wireframe and the finished component are slightly different. See:

qrcode

Id go with 2 `<Col md={6}>` and the elements should lie naturally within them then.

pawelpaszki:
ping @sedroche

ui/src/components/overview/BuildDownloadLinks.scss
```diff
@@ -0,0 +1,6 @@
+.build-download-component {
+  margin: 1em 0;
```
sedroche: I would remove this. We can set any layout in the parent component.

PR #28, opened 2018-07-31
[WIP] Create new mobile client UI

wojta:
https://issues.jboss.org/browse/AEROGEAR-7665

- [x] create mobile client form
- [x] make it modal dialog
- [ ] add validation
- [ ] prepare for backend connection

creationclient


ui/src/components/App.scss
```diff
@@ -0,0 +1,30 @@
+.platform-item > .platform-item-icon {
```
wojta: I think that this is not correct location for the styles. I should move it somewhere else... Not sure where?

ui/src/components/create_client/CreateXamarinClient.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Component for the Android specific create mobile client form.
+ */
+class CreateXamarinClient extends Component {
+    render() {
+        const formFields = [
+            {
+                controlId:'appName',
+                label:'* Application Name',
```
wojta: Is there some recommended proper way in Patternfly (bootstrap) to mark the field required?

ui/src/components/App.scss
ciaranRoche: At the moment if we need our own `sass` in a `js` file we add a `scss` file specifically for that `js`

wojta:
Moving to PR https://github.com/aerogear/mobile-client-service/pull/34 to use a local branch.

PR #29, opened 2018-07-31
Filter mobile services and list clients

sedroche:
- Adding list command for clients
- filtering `serviceInstances` which have been tagged `"mobile-service"`

ping @wei-lee 

**This branch now requires a new env var. Set:**

```.bash
export KUBERNETES_CONFIG=$HOME/.kube/config
```


Gopkg.toml
```diff
@@ -55,8 +55,7 @@ required = [
 
 [[override]]
   name = "k8s.io/client-go"
-  # revision for tag "kubernetes-1.10.0"
-  revision = "23781f4d6632d88e869066eaebb743857aa1ef9b"
+  version = "kubernetes-1.9.3"
```
wei-lee: this version of k8s isn't compatible with service-catalog client. That's why I updated it to 1.10.0. To get it work with the operator-sdk, you can update all the k8s.io libs to 1.10.0. That's what I am going to do in my branch anway.

pkg/web/mobileResourceHandler.go
```diff
@@ -0,0 +1,54 @@
...
+func (mrh *MobileResourceHandler) list(context echo.Context) error {
+	mobileClientResourceList, err := mrh.lister.List()
+	if err != nil {
+		return err
+	}
+
+	return context.JSON(http.StatusOK, mobileClientResourceList)
+}
+
+func (mrh *MobileResourceHandler) watch(context echo.Context, resourceVersion string) error {
```
wei-lee: looks like this doesn't do anything. Is it right?

Gopkg.toml
sedroche: will try this thanks

pkg/web/mobileResourceHandler.go
sedroche: Yeah it's not needed currently but we will need watch handling in the future. I can remove it until we need it if you want?

wei-lee: no, it's ok. We can leave it here and it will remind us what's required.

pkg/mobile/serviceInstances.go
```diff
@@ -17,5 +18,37 @@ func NewServiceInstanceLister(scClient scv1beta1.ServicecatalogV1beta1Interface)
...
+	getOpts := v1.GetOptions{}
+	serviceInstanceList := ServiceInstanceList{}
+
+	serviceInstances, err := lister.scClient.ServiceInstances(namespace).List(listOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, si := range serviceInstances.Items {
+		sc, err := lister.scClient.ClusterServiceClasses().Get(si.Spec.ClusterServiceClassRef.Name, getOpts)
```
wei-lee: am I right to say that the `mobile-service` tag is only available on the `ServiceClass`, but not on the `ServiceInstance`?

sedroche: Yeah it is only available on the `ServiceClass`.
There was a problem with tagging/labeling the `ServiceInstance` on creation

sedroche:
@ciaranRoche This look ok? https://github.com/aerogear/mobile-client-service/blob/45e87239634980cbd9ad112d72e2895854db95e2/README.md#run

pkg/mobile/serviceInstances.go
wei-lee: Right, ok. In this case you will need to update the test to add an instance of `ServiceClass`. The test doesn't have one ATM. That's why the build is failing.

README.md
```diff
@@ -39,6 +39,22 @@ If you don't have openshift running
...
+```
+
+Now set the `NAMESPACE` env var to point at the `target-project`
+```bash
+export NAMESPACE=<target-project>
+```
+
+Also set `KUBERNETES_CONFIG` to point to your `.kube/config`
+```bash
+export KUBERNETES_CONFIG=$HOME/.kube/config
```
wei-lee: There is already a cli flag created to set this value: https://github.com/aerogear/mobile-client-service/blob/master/cmd/api-server/main.go#L105. We should either use it as well, or remove it completely and just use the env var.

sedroche: The operator sdk needs `KUBERNETES_CONFIG` set but I can do
```.go
os.Setenv("KUBERNETES_CONFIG", kubeconfig)
```
using the flag value if its defined.

What is the syntax to pass namespace through `make serve`? I have always been using the env var

pkg/mobile/serviceInstances.go
sedroche: :+1: I'll update the test

cmd/api-server/main.go
```diff
@@ -75,7 +76,16 @@ func main() {
 		web.SetupMobileBuildConfigsRoute(apiGroup, mobileBuildConfigsHandler)
 	}
 
-        web.SetClientRoutes(apiGroup)
+	{
+		mobileResourceClient, _, err := k8sclient.GetResourceClient("mobile.k8s.io/v1alpha1", "MobileClient", namespace)
+		if err != nil {
```
camilamacedo86: Should not the linter complain about this point? Should it not be `if ( err != nil ) `

pkg/mobile/serviceInstances.go
```diff
@@ -17,5 +18,37 @@ func NewServiceInstanceLister(scClient scv1beta1.ServicecatalogV1beta1Interface)
...
+
+func isMobileService(sc *v1beta1.ClusterServiceClass) bool {
+	return contains(sc.Spec.Tags, "mobile-service")
+}
+
+func contains(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
```
camilamacedo86: Should not better here adopted names instead of a letter for the vars `a`, and `s`, and `e` in order to clarify its propose?  Something as which follows. 

```go 
func contains(arr []string, value string) bool {
      for _, arrValue := range arr {
		if arrValue == value {
			return true
		}

```

Also,

camilamacedo86: Also, should not better keep this kind of implementation in utils file ?

wei-lee: for a long program, yes. But for a short program like this, unless the code is very very hard to understand, then it is ok to use short var names. It is very common in Golang where short var names are used.

ui/src/components/overview/Overview.js
```diff
@@ -23,6 +26,15 @@ class Overview extends Component {
...
       });
+  
+    fetch(serviceInstancesUrl)
+      .then(response => response.json())
+      .then(result => {
+        this.setState({serviceInstances: result.items});
+      })
+      .catch(err => {
+        console.error('Fetch error: ', err)
+      });
```
camilamacedo86: Should it be not better if it is encapsulated? ( E.g `fetchServiceItems(..,)` ) Something as follows. 

```javascript 
/**
   * Add small desc
   */
  componentDidMount = () => {
    /**
     * Add small desc
     */
    function getClientItems() {
      fetch(listClientsUrl)
        .then(response => response.json())
        .then(result => {
          this.setState({mobileClients: result.items});
        })
        .catch(err => {
          console.error('Fetch error: ', err)
        });
    }

    /**
     * Add small desc
     */
    function getServiceItems() {
      fetch(serviceInstancesUrl)
        .then(response => response.json())
        .then(result => {
          this.setState({serviceInstances: result.items});
        })
        .catch(err => {
          console.error('Fetch error: ', err)
        });
    }

    getClientItems();
    getServiceItems();

  }


```

sedroche: There are some things that need improvements here alright.
I want to encapsulate these backend calls into their own module. You can view these changes as WIP.

I felt this piece of work was fine as is and a follow up after to finish the integrating

pkg/mobile/serviceInstances.go
sedroche: I'm not against a utils file but I tend to move functions into one as the need arises.

camilamacedo86: @wei-lee I found this [here](https://github.com/golang/go/wiki/CodeReviewComments#variable-names), _Variable names in Go should be short rather than long_. It is really interesting.  👍

sedroche: @camilamacedo86  :+1:

wei-lee: @camilamacedo86 thanks for sharing. Good read indeed.

ui/src/components/overview/Overview.js
camilamacedo86: In my point of view if we try to do improvements as this one for all PRs we will have a better quality code, and easier to keep maintained. Also, if we start to push ourselves more in this direction it will be natural and it will be done automatically. 

Because after doing the merges and the features are functional is hard to come back and to do improvements/re-factories like that.

sedroche: yeah I agree the front end code is out of place a little.
The purpose of this PR is the back end changes which is what I want to land. I don't want to start adding changes in this PR that will be followed up on further PRs. I can remove the front end changes totally if you wish?

camilamacedo86: I did not get. It was just a suggestion/recommendation. IMHO, just extract the logic and add comments as the example will be great. However, it is up to you and the repo maintainers.

wei-lee:
@sedroche I think this one is ok to merge now. I am not too concerned about the coverall reports

sedroche:
@wei-lee Just fixing up some last comments and ill merge

PR #36, opened 2018-08-03
make sure the service can be running inside an openshift cluster

wei-lee:


wei-lee:
@sedroche @StephenCoady mind taking a quick look?

PR #44, opened 2018-08-07
 🐛 make sure the cookies are sent to the backend when using fetch

wei-lee:
@sedroche adding the extra option fixed the problem for me.


wei-lee:
@StephenCoady can you please review this one?

PR #45, opened 2018-08-08
Aerogear 7668 mobile service view

secondsun:
Motivation
Porting the Mobile Service UI to mobile-client-service

What is done
UI and some basic APIs

What needs to be done
UI needs to be wired into the actual back end so it shows live data
Need to define a service "object" to share.
Probably UI fixes and tweaks


ui/src/components/mobileservices/MobileServiceView.js
```diff
@@ -0,0 +1,81 @@
...
+            setupText: 'Identity Management SDK setup'
+          }, {
+            serviceLogoUrl: 'https://avatars1.githubusercontent.com/u/3380462?s=200&v=4',
+            serviceName: 'Mobile Metrics',
+            serviceId: 'dh-metrics-apb-wqm5c',
+            serviceDescription: 'Installs a metrics service based on Prometheus and Grafana',
+            setupText: 'Mobile Metrics SDK setups',
+            configuration : [{key:"A property",value:"A Value"},{key:"Another property",value:"Another Value"}]
+          }]
+        };
```
wei-lee: @secondsun can you query the `<url of the service>/api/serviceinstances` endpoint to see if you have enough information to build this ui?

secondsun: Thanks, will do.

Is there a service object I can use?  If not do you mind if I add something that works like ScheduleService does here :  https://github.com/devnexus/devnexus-pwa/blob/master/script/Schedule.js#L13

wei-lee: @secondsun you mean this [DataService](https://github.com/aerogear/mobile-client-service/blob/master/ui/src/DataService.js)?

wei-lee:
@secondsun mind rebasing the PR so I can merge it?

PR #48, opened 2018-08-09
AEROGEAR-7748 - add build pipeline stage

ciaranRoche:
## JIRA
https://issues.jboss.org/browse/AEROGEAR-7748

## SCREENSHOT
build-status

## VERIFY
To verify have at least one mobile client created, mobile CI/CD provisioned and a build created.

## NOTE
It was decided to keep this basic for now as a webhook is needed to get the different pipeline stages.

### DIFF
--- a/ui/src/components/build/BuildPipelineStage.js
+++ b/ui/src/components/build/BuildPipelineStage.js
@@ -1,10 +1,24 @@
 import React, { Component } from 'react';
+import Moment from "react-moment";
+
+import BuildStatus from '../common/BuildStatus';
+
+import "./BuildPipelineStage.css";
 
 class BuildPipelineStage extends Component {
+    // TODO: add webhook or poll end point to set pipeline stages
     render () {
         return (
-            <div>
-                 <span>Pipeline Stage</span>
+            <div className="pipeline-stage">
+                 <div className="pipeline-stage-name">
+                    <b>Status : </b> {this.props.build.status.phase}
+                 </div>
+                 <div className="pipeline-stage-icon">
+                    <BuildStatus build={this.props.build}/>
+                 </div>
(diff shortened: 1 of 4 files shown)


ciaranRoche:
@jhellar if you wouldn't mind having a look if you are happy it might be worth merging into your current PR

ui/src/components/build/BuildPipelineStage.js
```diff
@@ -1,10 +1,24 @@
 import React, { Component } from 'react';
+import Moment from "react-moment";
+
+import BuildStatus from '../common/BuildStatus';
+
+import "./BuildPipelineStage.css";
 
 class BuildPipelineStage extends Component {
+    // TODO: add webhook or poll end point to set pipeline stages
```
wei-lee: is there a special endpoint we need to get the pipeline info, or is the current builds/buildconfigs endpoint is enough?

ciaranRoche: The json object returned from `/builds` contain a `phase` which appears to be the right information. Maybe @JameelB can add some insight into what they used previously.

JameelB: @wei-lee @ciaranRoche There's an [annotation](https://github.com/aerogear/origin-web-console/blob/aerogear-mcp/app/scripts/directives/buildPipeline.js#L25) in the build object called `jenkinsStatus` which contains information on the pipeline stages. This is then used for rendering the [pipeline info](https://github.com/aerogear/origin-web-console/blob/aerogear-mcp/app/views/directives/_build-pipeline-expanded.html#L32-L45) on the build page.

ciaranRoche: Awesome, thanks @JameelB

PR #52, opened 2018-08-09
allow to set the resync period via env var

wei-lee:


wei-lee:
@StephenCoady @secondsun mind taking a quick look?

PR #56, opened 2018-09-21
Added redux and eslint

jhellar:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-7922


pb82:
👀

ui/src/configureStore.js
```diff
@@ -0,0 +1,17 @@
+import { createStore, applyMiddleware } from 'redux'
+import thunkMiddleware from 'redux-thunk'
+import { createLogger } from 'redux-logger'
+import rootReducer from './reducers'
+
+let middleware = [ thunkMiddleware ]
+if (process.env.NODE_ENV !== 'production') {
```
pb82: Is `NODE_ENV` available in the frontend?

ui/src/containers/CreateClient.js
```diff
@@ -1,14 +1,28 @@
...
+  PLATFORM_CORDOVA,
+  PLATFORM_XAMARIN,
+  CREATE_CLIENT_TYPE,
+  CLIENT_CONFIGURATION,
+  CREATE_CLIENT_APP_ID,
+  CREATE_CLIENT_NAME,
+  WIZARD_SELECT_PLATFORM,
+  WIZARD_CREATION_RESULT,
+  WIZARD_CONFIGURE_CLIENT
+} from '../components/create_client//Constants'
```
pb82: double / in `create_client//Constants`

ui/src/containers/CreateClient.js
```diff
@@ -17,9 +31,29 @@ class CreateClient extends Component {
...
     constructor(props) {
         super(props)
-        this.baseState = this.state;
+        this.baseState = this.state
     }
 
+    componentDidUpdate() {
+      if (this.state.loading && !this.props.apps.isCreating) {
+        if (this.props.apps.createError) {
+          this.setState({
```
pb82: I wanted to point out the usage of `setState` first but i think it's ok here because the component is a Wizard and needs to maintain internal state.

wojta: That's a question if we it is better to change the code to have state completely in Redux.

ui/src/configureStore.js
jhellar: Yes, it is added automatically by react.

ui/src/reducers/resource.js
```diff
@@ -0,0 +1,61 @@
...
+      case actions.listFailure:
+        return {
+          ...state,
+          isFetching: false,
+          fetchError: action.error
+        }
+      default:
+    }
+  }
+  if (actions.createRequest) {
```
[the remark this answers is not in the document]
jhellar: This `if` checks if there are actions defined for create requests. If so it will check if such an action was dispatched. If there was a `else if`, reducers for create actions would not be called as every resource also has a list actions.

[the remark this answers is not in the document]
pb82: Ok, understand. Wondering if we really need the outer `if`. `actions.createRequest` could be undefined but that would be ok because `action.type` should never be undefined. WDYT?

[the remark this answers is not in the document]
jhellar: yes, ok, that makes sense, will fix it

jhellar:
@pb82 @wojta should be fixed now. I've also added eslint and ran auto-fix.

PR #59, opened 2018-09-25
feat: local development scripts

psturc:
### Motivation
https://issues.jboss.org/browse/AEROGEAR-7956

### Verification steps
# Clone this branch
# Follow the steps [here](https://github.com/aerogear/mobile-client-service/compare/AEROGEAR-7956?expand=1#diff-04c6e90faac2675aa89e2176d2eec7d8R82)
# Run `./scripts/development.sh`
# Go to `localhost:3000` and open developer console -> there should be no errors related to `cors` issues
# Do some changes to `.go` files -> files should be rebuilt & Go server should reload
# Do some changes to `.js` files -> node app should reload
# Kill the script (Ctrl + C) - there should be no error
# In terminal, run `ps` -> there should be no node or go processes running


ui/src/DataService.js
```diff
@@ -1,4 +1,8 @@
-const baseUrl = '/api';
+let baseUrl = 'api';
```
jhellar: Should not there stay `/api`?

psturc: good point, I overlooked this

PR #60, opened 2018-09-25
Added patternfly ESLint checking

wojta:
There are ESLint rules for patternfly provided here: 
https://github.com/patternfly/patternfly-react/tree/master/packages/eslint-plugin-patternfly-react

Added checking for them.

### DIFF
--- a/ui/.eslintrc.js
+++ b/ui/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+  root: true,
+  extends: ['plugin:patternfly-react/recommended']
+};
--- a/ui/.eslintrc.json
+++ b/ui/.eslintrc.json
@@ -9,5 +9,7 @@
     },
     "env": {
         "browser": true
-    }
+    },
+    "plugins": ["plugin:patternfly-react/recommended"]
+    
 }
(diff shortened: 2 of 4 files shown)


ui/.eslintrc.json
```diff
@@ -9,5 +9,7 @@
     },
     "env": {
         "browser": true
-    }
+    },
+    "plugins": ["plugin:patternfly-react/recommended"]
```
jhellar: This should not be in `plugins`, but in `extends` (bug in patternfly's docs). Also the order of `extends` items should be like this:
```
"extends": [
        "airbnb",
        "plugin:patternfly-react/recommended"
    ],
```

camilamacedo86:
Hi Folks, 

I understand that we should use [JavaScript Standard Style](https://standardjs.com/) instead of it. 

c/c @wei-lee @darahayes

ui/.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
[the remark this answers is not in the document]
wojta: ok, possibly, I was following that documentation on the github.

ui/.eslintrc.json
wojta: +1

wojta:
closing this because of #61

PR #61, opened 2018-09-25
Added patternfly ESLint checking

jhellar:


camilamacedo86:
@jhellar I understand that we should be using [JavaScript Standard Style](https://standardjs.com/) instead of it. 

c/c @darahayes

secondsun:
Do we have a decision from the mailing list accepting the patternfly rules?
When we added the google rules on the android sdk we ran it by the list first and took some feedback.

jhellar:
Replaced with https://github.com/aerogear/mobile-client-service/pull/66. Closing.

PR #70, opened 2018-10-03
Add openshiftMasterURL to list of builds

psturc:
### Motivation
https://issues.jboss.org/browse/AEROGEAR-7979

### How to verify
💁‍♂️  Feel free to use [my project on community cluster](https://community.skunkhenry.com/console/project/psturc-test/overview) (ping me for user/pass if you don't have it)
1. Clone this branch and follow instructions [from here](https://github.com/aerogear/mobile-client-service/blob/master/README.md#local-development)
1. Make sure the OpenShift project you are targeting has Mobile CI/CD provisioned and also has some existing builds
1. In MDC, go to Builds, click on Build and verify all links (`Build #x`) are pointing to the correct page in OpenShift Console (also check those in build history)

### DIFF
--- a/pkg/web/buildsHandler.go
+++ b/pkg/web/buildsHandler.go
@@ -1,10 +1,12 @@
 package web
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/aerogear/mobile-developer-console/pkg/mobile"
 	"github.com/labstack/echo"
+	buildV1 "github.com/openshift/api/build/v1"
 )
 
 type MobileBuildsHandler struct {
@@ -13,9 +15,14 @@ type MobileBuildsHandler struct {
 	openshiftMasterURL string
 }
 
-type response struct {
-	mobile.BuildList
-	OpenshiftMasterURL string `json:"openshiftMasterUrl,inline"`
+type extendedBuildList struct {
+	buildV1.BuildList
+	Items []extendedBuild `json:"items" protobuf:"bytes,2,rep,name=items"`
+}
+
+type extendedBuild struct {
+	buildV1.Build
+	BuildURL string `json:"buildUrl,inline"`
 }
 
 func NewMobileBuildsHandler(buildsCRUDL mobile.BuildCRUDL, namespace string, openshiftMasterURL string) *MobileBuildsHandler {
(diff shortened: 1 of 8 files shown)


pkg/web/buildsHandler.go
```diff
@@ -25,6 +32,9 @@ func (mbh *MobileBuildsHandler) List(c echo.Context) error {
 		c.Logger().Errorf("error listing builds %v", err)
 		return c.NoContent(http.StatusInternalServerError)
 	}
-
-	return c.JSON(http.StatusOK, builds)
+	response := &response{
+		BuildList:          *builds,
+		OpenshiftMasterURL: mbh.openshiftMasterURL,
```
wei-lee: rather than return the url on its own, I think maybe use it to construct the url of each build, and then add the url to each build as a separate field (like buildURL)? There are at least 2 benefits:

1. Cleaner API. Otherwise without some documentation, people wouldn't be able to understand why the master url is returned here.
2. Hide the logic to build the full url of the builds. This only needs to be done on the server and the clients don't need to care.

WDYT?

psturc: I think it sounds reasonable. This will definitely reduce amount of changes needed in UI code to reflect this, which I'm looking at right now. Thanks

psturc:
@wei-lee @jhellar PR is ready for E2E verification. 
Tests are failing because of missing metadata in [this test](https://github.com/aerogear/mobile-developer-console/blob/master/pkg/web/buildsHandler_test.go#L45-L48) because of [this line](https://github.com/aerogear/mobile-developer-console/pull/70/files#diff-6ac8b5769a96fc5fefbba92ea1539d12R42)

pkg/web/buildsHandler.go
```diff
@@ -1,21 +1,58 @@
...
-		namespace:   namespace,
+		buildsCRUDL:        buildsCRUDL,
+		namespace:          namespace,
+		openshiftMasterURL: openshiftMasterURL,
+	}
+}
+
+// Extend existing struct buildV1.BuildList so each build contains URL
+// pointing to build in OpenShift Console
+func extendBuildList(bl buildV1.BuildList, mbh MobileBuildsHandler) *extendedBuildList {
```
wei-lee: don't think we need to pass the `MobileBuildsHandler` object here. It's better to just pass the url and the namespace value as args explicitly here.

PR #71, opened 2018-10-03
WebSockets - keep UI updated

jhellar:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-7948

# Verification steps

- setup OpenShift with mobile-core
- deploy Mobile CI/CD service
- build and run MDC
- open more browser windows with MDC
- in one of them create mobile apps
- new mobile apps should appear in other windows without refresh
- create buildConfig in OpenShift
- it should appear in MDCs
- in one MDC instance trigger build
- it should appear in other windows
- as the build progresses, UI should be updated

### DIFF
--- a/pkg/mobile/mobileClient.go
+++ b/pkg/mobile/mobileClient.go
@@ -137,7 +137,7 @@ func (h *MobileHandler) RemoveWatcher(watcher MobileWatcher) {
 	m.Unlock()
 }
 
-func (r *MobileClientRepoImpl) Watch() (watch.Interface, error) {
+func (r *MobileClientRepoImpl) Watch() func() (watch.Interface, error) {
 	if handler == nil {
 		sdk.Watch("mobile.k8s.io/v1alpha1", "MobileClient", r.namespace, 0)
 		handler = NewMobileHandler()
@@ -147,5 +147,7 @@ func (r *MobileClientRepoImpl) Watch() (watch.Interface, error) {
 		events: make(chan watch.Event),
 	}
 	handler.AddWatcher(watcher)
-	return watcher, nil
+	return func() (watch.Interface, error) {
+		return watcher, nil
+	}
 }
--- a/pkg/mobile/mobileClient.go
+++ b/pkg/mobile/mobileClient.go
@@ -94,9 +94,7 @@ func (w MobileWatcher) Stop() {
 			}
 		}
 	}()
-	m.Lock()
 	handler.RemoveWatcher(w)
-	m.Unlock()
 	done <- struct{}{}
 }
 
@@ -105,7 +103,12 @@ func (w MobileWatcher) ResultChan() <-chan watch.Event {
 }
 
 type MobileHandler struct {
-	watchers []MobileWatcher
+	watchers map[MobileWatcher]struct{}
+}
+
+func NewMobileHandler() *MobileHandler {
+	watchers := make(map[MobileWatcher]struct{})
+	return &MobileHandler{watchers}
 }
 
 func (h MobileHandler) Handle(c context.Context, e sdk.Event) error {
@@ -115,8 +118,8 @@ func (h MobileHandler) Handle(c context.Context, e sdk.Event) error {
 
 func (h *MobileHandler) NotifyWatchers() {
 	m.Lock()
-	for index := 0; index < len(h.watchers); index++ {
-		h.watchers[index].events <- watch.Event{
+	for watcher := range h.watchers {
+		watcher.events <- watch.Event{
 			Type:   "MobileAppsEvent",
 			Object: nil,
 		}
@@ -125,23 +128,19 @@ func (h *MobileHandler) NotifyWatchers() {
 }
 
 func (h *MobileHandler) AddWatcher(watcher MobileWatcher) {
-	h.watchers = append(h.watchers, watcher)
+	h.watchers[watcher] = struct{}{}
 }
 
 func (h *MobileHandler) RemoveWatcher(watcher MobileWatcher) {
-	index := 0
-	for ; index < len(h.watchers); index++ {
-		if h.watchers[index] == watcher {
-			break
-		}
-	}
-	h.watchers = append(h.watchers[:index], h.watchers[index+1:]...)
+	m.Lock()
+	delete(h.watchers, watcher)
+	m.Unlock()
 }
 
 func (r *MobileClientRepoImpl) Watch() (watch.Interface, error) {
 	if handler == nil {
 		sdk.Watch("mobile.k8s.io/v1alpha1", "MobileClient", r.namespace, 0)
-		handler = &MobileHandler{}
+		handler = NewMobileHandler()
 		sdk.Handle(handler)
 	}
 	watcher := MobileWatcher{
--- a/pkg/mobile/mobileClient.go
+++ b/pkg/mobile/mobileClient.go
@@ -2,6 +2,8 @@ package mobile
 
 import (
 	"context"
+	"sync"
+	"time"
 
 	"github.com/aerogear/mobile-developer-console/pkg/apis/aerogear/v1alpha1"
 	"github.com/operator-framework/operator-sdk/pkg/sdk"
@@ -11,6 +13,7 @@ import (
 
 var (
 	handler *MobileHandler
+	m       sync.Mutex
 )
 
 type MobileClientRepoImpl struct {
@@ -79,7 +82,22 @@ type MobileWatcher struct {
 }
 
 func (w MobileWatcher) Stop() {
+	done := make(chan struct{})
+	go func() {
+		drain := true
+		for drain {
+			select {
+			case <-done:
+				drain = false
+			case <-w.events:
+			case <-time.After(1 * time.Second):
+			}
+		}
+	}()
+	m.Lock()
 	handler.RemoveWatcher(w)
+	m.Unlock()
+	done <- struct{}{}
 }
 
 func (w MobileWatcher) ResultChan() <-chan watch.Event {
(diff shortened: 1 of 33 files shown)


pkg/mobile/mobileClient.go
```diff
@@ -66,3 +73,49 @@ func (r *MobileClientRepoImpl) DeleteByName(name string) error {
...
+
+func (w MobileWatcher) Stop() {
+	close(w.events)
+	index := 0
+	for ; index < len(handler.watchers); index++ {
+		if handler.watchers[index] == w {
+			break
+		}
+	}
+	handler.watchers = append(handler.watchers[:index], handler.watchers[index+1:]...)
```
wei-lee: Probably better to move the few lines here to become the `removeWatcher` method of the MobileHandler

pkg/web/wsHandler.go
```diff
@@ -0,0 +1,59 @@
...
+		CheckOrigin: func(r *http.Request) bool {
+			return true
+		},
+	}
+)
+
+func ServeWS(context echo.Context, watchInterface watch.Interface) {
+	ws, err := upgrader.Upgrade(context.Response(), context.Request(), nil)
+	if err != nil {
+		return
```
wei-lee: why the `err` object is not returned here?

pkg/web/wsHandler.go
```diff
@@ -0,0 +1,59 @@
...
+	for {
+		select {
+		case <-events:
+			ws.SetWriteDeadline(time.Now().Add(writeWait))
+			if err := ws.WriteMessage(websocket.TextMessage, []byte("event")); err != nil {
+				return
+			}
+		case <-pingTicker.C:
+			ws.SetWriteDeadline(time.Now().Add(writeWait))
+			if err := ws.WriteMessage(websocket.PingMessage, []byte{}); err != nil {
```
wei-lee: should we log any of the errors?

pkg/web/wsHandler.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/gorilla/websocket"
+	"github.com/labstack/echo"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+const (
+	writeWait  = 10 * time.Second
+	pongWait   = 60 * time.Second
+	pingPeriod = (pongWait * 9) / 10
```
wei-lee: probably should consider making those configurable.

pkg/web/serviceInstancesHandler.go
```diff
@@ -20,10 +20,20 @@ func NewMobileServiceInstancesHandler(serviceInstanceLister mobile.ServiceInstan
...
 	return c.JSON(http.StatusOK, si)
 }
+
+func (msih *MobileServiceInstancesHandler) Watch(c echo.Context) error {
+	watchInterface, err := msih.serviceInstanceLister.Watch()
+	if err != nil {
+		return err
+	}
+
+	ServeWS(c, watchInterface)
```
wei-lee: I think `ServeWS` should return an error and it should returned to the client if there is a problem to establish the connection.

pkg/mobile/mobileClient.go
```diff
@@ -66,3 +73,49 @@ func (r *MobileClientRepoImpl) DeleteByName(name string) error {
...
+			Type:   "MobileAppsEvent",
+			Object: nil,
+		}
+	}
+	return nil
+}
+
+func (r *MobileClientRepoImpl) Watch() (watch.Interface, error) {
+	if handler == nil {
+		sdk.Watch("mobile.k8s.io/v1alpha1", "MobileClient", r.namespace, 0)
```
wei-lee: what will happen when the last argument is set to `0`? How often will the `handler` being invoked?

pkg/mobile/mobileClient.go
```diff
@@ -66,3 +73,49 @@ func (r *MobileClientRepoImpl) DeleteByName(name string) error {
...
+	watchers []MobileWatcher
+}
+
+func (h MobileHandler) Handle(c context.Context, e sdk.Event) error {
+	for index := 0; index < len(h.watchers); index++ {
+		h.watchers[index].events <- watch.Event{
+			Type:   "MobileAppsEvent",
+			Object: nil,
+		}
+	}
```
wei-lee: probably better to move the few lines of code to become the `notifyWatchers` method of the `MobileHandler`

pkg/web/wsHandler.go
jhellar: ok, if we want to make this configurable, how best to go about it? optional command line arguments?

pkg/mobile/mobileClient.go
wei-lee: right, so it will only send event when there is a change, is that right?

jhellar: yes, that is how I understand it

pkg/web/wsHandler.go
wei-lee: yes, and environment variables

pkg/mobile/mobileClient.go
wei-lee: cool. thanks

jhellar:
@wei-lee Thanks for review, all comments should be addressed now.

pkg/mobile/mobileClient.go
```diff
@@ -66,3 +73,56 @@ func (r *MobileClientRepoImpl) DeleteByName(name string) error {
...
+
+func (w MobileWatcher) ResultChan() <-chan watch.Event {
+	return w.events
+}
+
+type MobileHandler struct {
+	watchers []MobileWatcher
+}
+
+func (h MobileHandler) Handle(c context.Context, e sdk.Event) error {
```
wei-lee: we should be able to add some unit tests for `MobileHandler`

jhellar:
@wei-lee I've added the unit test. While writing it I found data race issue which should also be fixed now.

pkg/mobile/mobileClient.go
```diff
@@ -66,3 +76,73 @@ func (r *MobileClientRepoImpl) DeleteByName(name string) error {
...
+			case <-done:
+				drain = false
+			case <-w.events:
+			case <-time.After(1 * time.Second):
+			}
+		}
+	}()
+	m.Lock()
+	handler.RemoveWatcher(w)
+	m.Unlock()
```
wei-lee: is this because a race condition will happen when the "Stop" method is called on multiple watchers? if so would a map a better option to store the watchers? This way it will be more efficient as you don't need to loop through the array to find the watcher and build a new array.

pkg/mobile/mobileClient_test.go
```diff
@@ -0,0 +1,62 @@
...
+	t.Run("test mobile handler", func(t *testing.T) {
+		handler = &MobileHandler{}
+		watcher1 := MobileWatcher{events: make(chan watch.Event)}
+		watcher2 := MobileWatcher{events: make(chan watch.Event)}
+		handler.watchers = append(handler.watchers, watcher1)
+		handler.watchers = append(handler.watchers, watcher2)
+		go func() {
+			for index := 0; index < 2; index++ {
+				handler.Handle(nil, sdk.Event{Object: nil, Deleted: false})
+			}
```
wei-lee: why call the handler twice here?

pkg/mobile/mobileClient_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+func TestMobileHandler(t *testing.T) {
+	t.Run("test mobile handler", func(t *testing.T) {
+		handler = &MobileHandler{}
+		watcher1 := MobileWatcher{events: make(chan watch.Event)}
+		watcher2 := MobileWatcher{events: make(chan watch.Event)}
+		handler.watchers = append(handler.watchers, watcher1)
+		handler.watchers = append(handler.watchers, watcher2)
```
wei-lee: It's possible to forget assign the array back in this way. So given that there is a "RemoveWatcher" method now, maybe worth moving this into a `AddWatcher` method?

pkg/mobile/mobileClient.go
jhellar: The race condition will probably happen also for multiple `Stop`s. I saw the issue when stopping one watcher and sending notifications in same time. Basically any read/write operations happening simultaneously will cause it. So I think that map would not help here.

jhellar:
@wei-lee it should be fixed now

pkg/mobile/mobileClient.go
wei-lee: A normal golang map won't help, but there is a special type of [Map in the sync package](https://golang.org/pkg/sync/#Map) which will help. Probably worth taking a look.

pkg/mobile/mobileClient_test.go
jhellar: I just wanted to verify, that
- when there are watchers registered, they all receive notification
- and when watcher unregisters [1], it will not receive more notifications, but the rest of the watchers will do

[1] https://github.com/aerogear/mobile-developer-console/pull/71/files#diff-b5c34eb680cd64a7334e27350c53c584R36

pkg/mobile/mobileClient.go
jhellar: thanks, will take a look

pkg/mobile/mobileClient_test.go
wei-lee: right, now I get it. Might worth adding some comments to explain it. Thanks.

jhellar:
@wei-lee I've tried few things with sync.map:

- changing type of keys in sync.map from `MobileWatcher` to `*MobileWatcher` - that blocks the execution as the watcher is deleted from the map, but there is still pending notification in channel with no consumer

- so I tried to close the channel -> panic

- I’ve tried to recover from the panic, but here is another problem - when the function recovers it will not continue from the next command after the panicking one (according to https://blog.golang.org/defer-panic-and-recover it should, but done some testing in playground and it is not), so that means other watchers in map won’t be notified -> tests are failing (will not finish - goroutine is blocked)

- because of that recovering from panic would probably lead to some complex code (I don’t even know now how to properly handle it)

- having watchers as values of map instead of keys would not be useful for the delete operation - reason for using map

- so maybe using normal map (not the sync one) with the mutexes? - I've pushed another commit for this

wei-lee:
@jhellar thanks for the details analysis. Agreed that for now it's best to use a map and a lock to achieve what we want.

PR #86, opened 2018-10-12
add route for user info

wei-lee:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7992

## What
Show the current logged in user name in the console

## How

The openshift oauth proxy will forward the username & email info in the request headers. We just need to retrieve them and return back to the client.

## Verification Steps

1. Provision the MDC from the service catalog. Make sure the oauth proxy provisioned to protected the service.
2. Login to the service. You should see your name show up in the top right corner.

## Checklist:


psturc:
👀

PR #87, opened 2018-10-12
Aerogear 7912 service binding

secondsun:
This is the work in progress for creating service bindings

What's (mostly) done : 
 Lots of UI stuff
 Some server stuff

What's (mostly) missing
 Actually creating and deleting the binding

How to create and delete the binding : 
So I got blocked on how you actually create the binding.  Specifically I know in do I need to call Create on the api client and then add the form data to the secret that gets created by the binding.  But I ran out of time to bash my head against the wall hard enough to actually figure out what parameters need to be set and what parameters OpenShift can provide itself.  

Whoever wants to pick this up and finish it next week can, otherwise I'll pick it back up after PTO.


secondsun:
ping @laurafitzgerald Here is the PR for my work in progress.

secondsun:
@wei-lee Heya so when I started I don't think we had started using redux, so the create and maybe some of the fetch code can use a tidy.  Need to add a unit test or three to make coveralls happy.

pkg/mobile/bindableServices.go
```diff
@@ -0,0 +1,189 @@
...
+
+		bindableService := BindableMobileService{}
+
+		csc, err := lister.scClient.ClusterServiceClasses().Get(serviceInstance.Spec.ClusterServiceClassRef.Name, getOpts)
+
+		if err != nil {
+			return nil, err
+		}
+
+		if csc.Spec.ExternalMetadata != nil {
```
wei-lee: might be better if we move this into its own function to reduce the size of this one.

pkg/mobile/bindableServices.go
```diff
@@ -0,0 +1,189 @@
...
+
+		for _, sb := range servicePlans.Items {
+			if sb.Spec.ClusterServiceClassRef.Name == csc.Name {
+				servicePlan = sb
+			}
+		}
+
+		var serviceBinding ServiceBinding
+
+		for _, sb := range serviceBindings.Items {
```
wei-lee: similarly, I think this can be moved to its own function to reduce the size of this one

pkg/web/serviceBindingsHandler.go
```diff
@@ -0,0 +1,95 @@
...
+		namespace:            namespace,
+	}
+}
+
+func (msih *BindableMobileServiceHandler) List(c echo.Context) error {
+	mobileClientName := c.Param("name")
+	si, err := msih.bindableServiceCRUDL.List(msih.namespace, mobileClientName)
+	if err != nil {
+		c.Logger().Errorf("error listing service bindings %v", err)
+		return c.NoContent(http.StatusInternalServerError)
```
wei-lee: we now need to return the error code and the error message. See #102

ui/src/components/mobileservices/MobileServiceView.js
```diff
@@ -1,46 +1,71 @@
...
-      }, {
-        serviceLogoUrl: 'https://avatars1.githubusercontent.com/u/3380462?s=200&v=4',
-        serviceName: 'Mobile Metrics',
-        serviceId: 'dh-metrics-apb-wqm5c',
-        serviceDescription: 'Installs a metrics service based on Prometheus and Grafana',
-        setupText: 'Mobile Metrics SDK setups',
-        configuration: [{ key: 'A property', value: 'A Value' }, { key: 'Another property', value: 'Another Value' }],
-      }],
+      boundServices: [],
+      unboundServices: [],
```
wei-lee: this is where using Redux will help. You don't need to manually update the state in this component.

secondsun:
@wei-lee Yeah.  So here's my plan of attack
1.  Port to redux
2. refactor the go methods to make them xmaller
3. add tests
4. Turn the filtering bug you found into a JIRA
5. Add JIRAs for updating the forms for the APBs (each one may be two, one for the apb data and one for the ui update)
6. Add a Jira to get live updating working (ie having the UI refresh when services are bound/unbound)
7. Add a jira to make the delete dialog behave (it doesn't close when you click delete)

ui/src/DataService.js
```diff
@@ -88,7 +89,34 @@ const dataService = {
     }
     return response.json();
   },
+  createBinding: async (mobileClientName, serviceInstanceName, credentialSecretName, parametersSecretName, serviceClassExternalName, formData) => {
+
+    var binding = {
```
wojta: let/const

ui/src/components/bindingUtils.js
```diff
@@ -0,0 +1,19 @@
+
+export const DNS1123_SUBDOMAIN_VALIDATION = {
+    pattern: /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/,
+    maxlength: 253,
+    description: 'Name must consist of lower-case letters, numbers, periods, and hyphens. It must start and end with a letter or a number.'
+  };
+
+export function createSecretName(prefix) {
+    var secretNamePrefixMaxLength = DNS1123_SUBDOMAIN_VALIDATION.length - 6; //We append a 5 digit code and a -;
```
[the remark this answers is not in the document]
wojta: shouldn't we use `const` here

ui/src/components/bindingUtils.js
```diff
@@ -0,0 +1,19 @@
...
+
+    if (prefix.length > secretNamePrefixMaxLength) {
+        prefix = prefix.substring(0, secretNamePrefixMaxLength);
+    }
+
+    var randomString = Math.round((Math.pow(36, 6) - Math.random() * Math.pow(36, 6))).toString(36).slice(1);
```
wojta: const

wei-lee:
@secondsun sounds like a good plan to me

secondsun:
@wei-lee Got the changes in for everything but testing.  If you approve, minus the testing, want to merge it and add a card to the jira for writing tests?  I know we wanted to have a big eslint autofix at the end of the sprint (wednesday)

wei-lee:
@secondsun sounds good to me. Do you mind creating the jira? Thanks.

PR #89, opened 2018-10-15
Remove the mock builds data in the card view

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7996

## What
Removed the mock builds data in the card view

## Verification Steps
1. Create few apps with few build configs
2. Verify that there is no build info in the app cards
3. Start some builds, make some of them fail by providing wrong git source url (to the build config)
4. Verify build statuses

## Checklist:

### DIFF
--- a/ui/src/components/overview/MobileClientCardView.js
+++ b/ui/src/components/overview/MobileClientCardView.js
@@ -14,8 +14,6 @@ import CreateClient from '../../containers/CreateClient';
 
 const mockServices = [{ type: 'metrics' }, { type: 'keycloak' }, { type: 'sync' }, { type: 'push' }];
 
-const mockBuilds = [{ status: 'success' }, { status: 'failed' }, { status: 'success' }, { status: 'failed' }];
-
 class MobileClientCardView extends Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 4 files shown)


ui/src/components/overview/MobileClientCardView.js
```diff
@@ -59,6 +57,34 @@ class MobileClientCardView extends Component {
...
+      }
+      return prev;
+    }, {});
+    let numFailedBuilds = 0;
+    for (const build in lastBuilds) {
+      if (lastBuilds[build].buildStatus === 'Failed') {
+        numFailedBuilds++;
+      }
+    }
+    const numInProgressBuilds = builds.filter(build => build.status.phase === 'Running').length;
```
wei-lee: why the in progress number of builds is calculated from all the builds, but the failed number of builds is not? should we keep the behaviour consistent?

wei-lee:
@jhellar is there a cluster I can verify this change with? I don't have the CI/CD setup locally.

ui/src/components/overview/MobileClientCardView.js
jhellar: The reason for showing failed builds only for last builds is because that is what is useful for user IMO. If we would show all failed builds, this number would just grow with time and there would be no value in it. On the other hand, with in-progress builds, I thing there is a value in knowing how many builds are in progress, even if there are more for one build config. WDYT?

PR #92, opened 2018-10-16
Link to documentation in navbar

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7998

### DIFF
--- a/ui/src/containers/Header.js
+++ b/ui/src/containers/Header.js
@@ -12,6 +12,7 @@ class Header extends Component {
     const {user} = this.props;
     const helpDropdowns = [{
       text: "Documentation",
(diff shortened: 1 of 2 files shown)


ui/src/containers/Header.js
```diff
@@ -12,9 +12,7 @@ class Header extends Component {
     const {user} = this.props;
     const helpDropdowns = [{
       text: "Documentation",
-      onSelect: () => {
-        console.log("Implement me!!");
-      }
+      href: 'http://docs.aerogear.org/'
```
psturc: Should we add a comment here that once we've got documentation for MDC, the link should navigate you straight to MDC service?

jhellar: Makes sense, will add it.

PR #96, opened 2018-10-18
Update app create dialog to reference style

ziccardi:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-7997/

## What

- Replaced the wizard with a modal form
- Added app name and app id validation
- Added help in case of bad input
- Added a component to manage a list of mutually exclusive platform to be selected (PlatformItems)
- Enhanched the PlatformItem component to be able to self configure itself based on the passed in platform type (image/icon, etc.)
- Enhanched the CreateClient component to be able to receive as property the list of platform to be supported
- Enhanched the CreateClientBaseClass: now subclasses can easily add custom fields, custom validataion, custom labels, etc.
- Changed layout to reflect the provided [flat](https://redhat.invisionapp.com/share/Z2O2VWUU9J8#/screens/320204113)
- Replaced react state management with react-redux

### Start Editing
Start Editing

### Valid Input
Valid Input

### Invalid Input
Invalid Input

### Error from the server (to get this I disabled client side validation)
Error from the server

## Checklist:

### DIFF
--- a/ui/src/actions/apps.js
+++ b/ui/src/actions/apps.js
@@ -37,32 +37,57 @@ export const deleteApp = name => fetchAction(
   async () => DataService.deleteApp(name),
 )();
 
+// CREATE CLIENT APP DIALOG ACTIONS
 export const APP_PLATFORM_REGISTER = 'PLATFORM_REGISTER';
 export const APP_PLATFORM_SELECT = 'PLATFORM_SELECT';
 export const APP_FORM_SETSTATUS = 'FORM_SETSTATUS';
 export const APP_FORM_RESET = 'FORM_RESET';
 export const APP_FIELD_SETVALUE = 'FIELD_SETVALUE';
 
+/**
+ * The list of platforms that the user can choose from are dynamic.
+ * This method is called as soon as the list of the platform to be supported is know
+ * to register into the state an initial state for all of them.
+ * @param {string} platform 
+ */
 export const registerPlatform = (platform) => ({
   type: APP_PLATFORM_REGISTER,
-  platform: { name: platform.name } 
+  platform: { name: platform.name }
 })
 
+/**
+ * When a platform is selected, this action is called to update the state with the selection
+ * and deselect all the other platforms.
+ * @param {string} platform 
+ */
 export const selectPlatform = (platform) => ({
   type: APP_PLATFORM_SELECT,
-  platform: { name: platform } 
+  platform: { name: platform }
 })
 
+/**
+ * This action is called to update the state with the current validation state of the form.
+ * @param {boolean} newStatus 
+ */
 export const setStatus = (newStatus) => ({
   type: APP_FORM_SETSTATUS,
-  payload: { status: newStatus}
+  payload: { status: newStatus }
 })
 
+/**
+ * Sets the value of a field into the state
+ * @param {string} fieldId id of the field
+ * @param {string} value the value of the field
+ * @param {boolean} isValid if the field is valid or not
+ */
 export const setFieldValue = (fieldId, value, isValid) => ({
   type: APP_FIELD_SETVALUE,
-  payload: { name: fieldId, value: value, valid: isValid}
+  payload: { name: fieldId, value: value, valid: isValid }
 })
 
+/**
+ * Resets the form to its initial state.
+ */
 export const resetForm = () => ({
   type: APP_FORM_RESET
 })
--- a/ui/src/reducers/resource.js
+++ b/ui/src/reducers/resource.js
@@ -1,15 +1,10 @@
-import { 
-  APP_PLATFORM_REGISTER, 
-  APP_PLATFORM_SELECT, 
+import {
+  APP_PLATFORM_REGISTER,
+  APP_PLATFORM_SELECT,
   APP_FORM_SETSTATUS,
   APP_FORM_RESET,
-  APP_FIELD_SETVALUE } from "../actions/apps";
-
-// export const APP_PLATFORM_REGISTER = 'platform/REGISTER';
-// export const APP_PLATFORM_SELECT = 'platform/SELECT';
-// export const APP_FORM_SETSTATUS = 'form/SETSTATUS';
-// export const APP_FORM_RESET = 'form/RESET';
-// export const APP_FIELD_SETVALUE = 'field/SETVALUE';
+  APP_FIELD_SETVALUE
+} from "../actions/apps";
 
 const defaultState = {
   isFetching: false,
@@ -149,20 +144,25 @@ const resourceReducer = actions => (state = defaultState, action) => {
   }
 };
 
-
+/**
+ * Reducers for the create client app dialog.
+ * @param {string} state 
+ * @param {*} action 
+ */
 function createClientAppDialog(state, action) {
   switch (action.type) {
-    case APP_FORM_RESET: 
-      var newCreateClientAppDialog= {
-        platforms: state.createClientAppDialog.platforms,
-        fields: {}
-      }
-
-      return { ...state, createClientAppDialog: newCreateClientAppDialog};
+    case APP_FORM_RESET:
+      return {
+        ...state,
+        createClientAppDialog: {
+          platforms: state.createClientAppDialog.platforms,
+          fields: {}
+        }
+      };
     case APP_PLATFORM_REGISTER:
-      var newState = {...state};
+      var newState = { ...state };
       if (!newState.createClientAppDialog.platforms[action.platform.name]) {
-        newState.createClientAppDialog.platforms[action.platform.name] = { selected: false};
+        newState.createClientAppDialog.platforms[action.platform.name] = { selected: false };
       }
       return newState;
     case APP_PLATFORM_SELECT:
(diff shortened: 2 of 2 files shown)


ui/src/components/create_client/VerticalFormField.js
```diff
@@ -0,0 +1,52 @@
...
+        {showHelp && help && <Form.HelpBlock>{help}</Form.HelpBlock>}
+      </Form.FormGroup>
+    );
+  }
+  return (
+    <Form.FormGroup {...formGroupProps}>
+      {label && <Form.ControlLabel>{label}</Form.ControlLabel>}
+      {formControl(controlProps)}
+      {showHelp && help && <Form.HelpBlock>{help}</Form.HelpBlock>}
+    </Form.FormGroup>
```
wei-lee: I think you can move the if statement inside the jsx so that you don't need the duplicated code:

```suggestion
   const helpContorl = (...);
   <Form.FormGroup {...formGroupProps}>
      {label && <Form.ControlLabel>{label}</Form.ControlLabel>}
      {useFieldLevelHelp && helpContorl}
      {formControl(controlProps)}
      {showHelp && help && <Form.HelpBlock>{help}</Form.HelpBlock>}
    </Form.FormGroup>
```

ui/src/containers/CreateClient.js
```diff
@@ -1,217 +1,162 @@
...
-  WIZARD_SELECT_PLATFORM,
-  WIZARD_CREATION_RESULT,
-  WIZARD_CONFIGURE_CLIENT,
 } from '../components/create_client/Constants';
-import { createApp } from '../actions/apps';
-import '../components/create_client/create_client.css';
+import CreateAndroidClient from '../components/create_client/CreateAndroidClient';
+import CreateCordovaClient from '../components/create_client/CreateCordovaClient';
+import CreateXamarinClient from '../components/create_client/CreateXamarinClient';
+import CreateIOSClient from '../components/create_client/CreateIOSClient';
```
wei-lee: it's likely that we need to make those platforms configurable. If it's not too much trouble, can we make them became part of props of the component, so that we can easily change them, ideally in just 1 place (the top level view)

ui/src/containers/CreateClient.js
```diff
@@ -1,217 +1,162 @@
...
+        </ControlLabel>
+        <PlatformItems itemSelected={this.selectPlatform} >
+          <PlatformItem title="Android" id="platform_android" inclass="fa fa-android"/>
+          <PlatformItem title="Cordova" id="platform_cordova">
+            <span><img src="../../img/cordova.png" alt="Cordova" /></span>
+          </PlatformItem>
+          <PlatformItem title="iOS" id="platform_ios" inclass="fa fa-apple" />
+          <PlatformItem title="Xamarin" id="platform_xamarin" >
+            <span><img src="../../img/xamarin.svg" alt="Xamarin" /></span>
+          </PlatformItem>
```
wei-lee: similar let's try make the platforms configurable.

ui/src/containers/CreateClient.js
```diff
@@ -1,217 +1,162 @@
...
+          Application Platform
+        </ControlLabel>
+        <PlatformItems itemSelected={this.selectPlatform} >
+          <PlatformItem title="Android" id="platform_android" inclass="fa fa-android"/>
+          <PlatformItem title="Cordova" id="platform_cordova">
+            <span><img src="../../img/cordova.png" alt="Cordova" /></span>
+          </PlatformItem>
+          <PlatformItem title="iOS" id="platform_ios" inclass="fa fa-apple" />
+          <PlatformItem title="Xamarin" id="platform_xamarin" >
+            <span><img src="../../img/xamarin.svg" alt="Xamarin" /></span>
```
wei-lee: can we maybe move the logic of either using the `fa` icon or image to the `PlatformItem` itself, so that when it's been used by other components, all that needed to be provided is the name of the platform.

jhellar:
The flats do not contain `package name` field. Do we still need it?

jhellar:
Can you adjust styling, so that there is not the underline for platform names:
underline

jhellar:
In browser console there are two warnings:

- Warning: Each child in an array or iterator should have a unique "key" prop.
- Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.

ziccardi:
> The flats do not contain `package name` field. Do we still need it?

👍  Tried to create an app without it and worked, so looks like it is not needed anymore. Removing!

ziccardi:
@wei-lee @jhellar : mind giving a look? I should have addressed all your comments.

wei-lee:
@ziccardi I am still having the problem I commented previously: the values in the app name and package name fields are cleared when I change the value of the platform.

ui/src/components/create_client/CreateMobileClientBaseClass.js
```diff
@@ -0,0 +1,124 @@
...
+    this.state = {
+      valid: false,
+      validationState: {}
+    }
+    this.config = {
+      platform: platformName,
+      appName: {
+        label: '* App Name',
+        example: 'myapp',
+        help: 'App name must match ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
```
wei-lee: while this is the rule we are using to validate the value, I don't think we should present it in the help message. I think a better help message should be like "app name should only contain letters, numbers and dots".

ui/src/components/create_client/CreateMobileClientBaseClass.js
```diff
@@ -0,0 +1,124 @@
...
+      },
+      appIdentifier: {
+        label: '* Package Name',
+        example: 'org.aerogear.myapp',
+        help: 'Package name must match ^[a-zA-Z][\\w]*(\\.[a-zA-Z][\\w]*)+$',
```
wei-lee: the same here.

ui/src/components/overview/MobileClientCardView.js
```diff
@@ -37,7 +43,7 @@ class MobileClientCardView extends Component {
       <EmptyState>
         <EmptyStateTitle>{this.emptyStateMessage.noAppsCreated}</EmptyStateTitle>
         <EmptyStateAction>
-          <CreateClient createButtonSize="large" />
+          <CreateClient createButtonSize="large" platforms={[PLATFORM_ANDROID, PLATFORM_CORDOVA, PLATFORM_IOS, PLATFORM_XAMARIN]}/>
```
wei-lee: let's move the platforms props to be part of the `App` and pass it down here.

ui/src/reducers/resource.js
```diff
@@ -128,8 +131,44 @@ const resourceReducer = actions => (state = defaultState, action) => {
...
+      return createClientAppDialog(state, action);
   }
 };
 
+
+function createClientAppDialog(state, action) {
+  switch (action.type) {
+    case 'platform/REGISTER':
+      var newState = {...state};
+      newState.createClientAppDialog.platforms[action.platform.name] = { selected: false};
```
jhellar: This would change the previous state. It should look something like this:
```
var newState = {
  ...state,
  createClientAppDialog: {
    ...state.createClientAppDialog,
    platforms: {
      ...state.createClientAppDialog.platforms,
      [action.platform.name] = { selected: false }
    }
  }
};
```

ui/src/actions/apps.js
```diff
@@ -36,3 +36,23 @@ export const deleteApp = name => fetchAction(
   [APP_DELETE_REQUEST, APP_DELETE_SUCCESS, APP_DELETE_FAILURE],
   async () => DataService.deleteApp(name),
 )();
+
+export const registerPlatform = (platform) => ({
+  type: 'platform/REGISTER',
+  platform: { name: platform.name, selected: platform.selected }
```
jhellar: Is `selected` needed here? It is not used in reducer [1].

[1] https://github.com/aerogear/mobile-developer-console/pull/96/commits/11e2eb36803253cf2ddebd3f9e0a51bafb699525#diff-a7113868962bf92abdf2145b05c18bbfR143

ui/src/components/create_client/CreateMobileClientBaseClass.js
ziccardi: I agree, however here it is a bit more complicated than that.
The appIpdentifier

- Must have at least two segments (one or more dots).
- Each segment must start with a letter.
- All characters must be alphanumeric or an underscore [a-zA-Z0-9_].

ui/src/actions/apps.js
```diff
@@ -36,3 +36,27 @@ export const deleteApp = name => fetchAction(
   [APP_DELETE_REQUEST, APP_DELETE_SUCCESS, APP_DELETE_FAILURE],
   async () => DataService.deleteApp(name),
 )();
+
+export const registerPlatform = (platform) => ({
+  type: 'platform/REGISTER',
```
wei-lee: a very small thing: can we follow the same convention of types in this file, use "_" instead of "/"?

PR #98, opened 2018-10-23
 ✨allow enable/disable the build tab via environment variables

wei-lee:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-8007

## What

To allow enable/disable the build tab via environment variables.

## Why

See JIRA

## How

1. The server will use the environment variable to enable/disable build-related routes
2. A new endpoint is added to the server to return the value of the environment variable for UI
3. The UI will load the env var from server on load.

## Verification Steps

1. Checkout the PR
2. set the "ENABLE_BUILD_TAB" to env var to true, and start the console and the server.
3. Check the UI and you should see the build tab. If you look at the network requests and you should see there are requests made for builds and buildconfigs.
4. set the env var to false and repeat step 3, this time you should not see the build tab, and there should be no builds/buildconfigs requests made to the backend.
5. You should also see some of the requests for the build-related endpoints are getting 404 in the network inspector.
 
## Checklist:

- [] Changes have been successfully verified by another team member 

## Progress

- [x] Use env var to control the build-related routes on the server 
- [x] Pass the env var config to the front end UI
- [x] Hide/show the build-related UI components based on the env var

## Additional Notes

1. Originally I was planning to render the `index.html` page on the Golang server, and we can have a placeholder in the `index.html` page and it will be replaced with server configuration data when it's been rendered. However, this turns out doesn't work in the development mode: `create-react-app` is running a local nodejs server to serve the content of the static files, and the request for `index.html` file never reaches the Golang backend (more details can be found [here](https://github.com/facebook/create-react-app/issues/1703)). So I ended up creating a new endpoint to return a generated script that will set the server data, and this script will be loaded before the bundled js file is loaded. This will work for both local development and server. The only downside is that this file will not be cached by the serviceworker. But since supporting offline is a no-goal anyway, it's not a problem for us. We may review this in the future if there is a better way to do this.
2. It also turns out that `create-react-app` has a http proxy server built-in, all we have to do is to provide the configuration in the package.json file. So I updated the settings for the `proxy` field and removed the CORS server implementation.


wei-lee:
@ziccardi @psturc @wojta @jhellar @secondsun this PR is ready for review now.

wei-lee:
@psturc ah, yes. The instructions were not updated correctly. There should be no requests made for the builds-related endpoints anymore.

PR #100, opened 2018-10-25
[WIP] Client details screen redesigned according to flats

wojta:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8031

## What
Changing UI screen to look like on https://redhat.invisionapp.com/share/Z2O2VWUU9J8#/screens/321837764

## How
Changing CSS styles, creating few placeholders

## Verification Steps

1. Deploy app configured against some OpenShift
2. Start with `make serve` 
3. Go to `http://localhost:4000`
4. Click client app to go to the client details screen

## Checklist:

## Progress

- [x] Basic layout and placeholders
- [ ] Code editor line numbers

## Additional Notes
 
image


ui/src/DataService.js
```diff
@@ -13,7 +13,12 @@ const getWSUrl = () => {
 let baseUrl = '/api';
 let wsUrl = `${getWSUrl()}/api`;
 
-const fetchItems = async (url) => {
+if (process.env.NODE_ENV !== 'production') {
+  baseUrl = 'http://localhost:8080/http://localhost:4000/api';
+  wsUrl = 'ws://localhost:4000/api';
+}
```
wei-lee: this should be removed.

ui/src/components/common/CopyToClipboardMultiline.js
```diff
@@ -1,36 +1,35 @@
...
-    );
-  }
-}
+const spanizeString = code =>
+  code.split('\n').reduce(
+    (accu, line) => (
+      <React.Fragment>
+        {accu}
+        <span>{line}</span>
+      </React.Fragment>
```
wei-lee: is it necessary to return the outer `React.Fragment` here?

wei-lee: I actually can't get MDC to work with this settings.

wojta: It's how the React works with the DOM, I think. I don't know if there is other option. I'll check it with @jhellar.

ui/src/DataService.js
wojta: ok, probably merge mishap, I'll take it from master.

ui/src/components/common/CopyToClipboardMultiline.js
jhellar: @wei-lee @wojta works for me. IMO the problem is that jsx won't allow for more than one root tag in one expression - removing `<React.Fragment>` here would not work. But there might be another problem - reduce in this way with every line will wrap the previous one in another fragment. IMO it should not be a problem even for many lines. But there is other possible solution, which won't do the wrapping:
```
const spanizeString = code => (
  <React.Fragment>
    {code.split('\n').map((line, index) => (
      <span key={index}>{line}</span>
    ))}
  </React.Fragment>
);
```

ui/src/components/common/CopyToClipboardMultiline.scss
```diff
@@ -1,19 +1,27 @@
-.copy-to-clipboard-multiline {
-  display: block;
-  overflow-x: auto
-}
+pre {
```
jhellar: IMO we should keep `.copy-to-clipboard-multiline` (i.e. `.copy-to-clipboard-multiline pre {...}`) here as well, otherwise this styling would be applied to every `pre`.

wojta: ok

wojta:
All fixed now.

ui/src/components/common/CopyToClipboardMultiline.js
wojta: thanks @jhellar , I used your code.

PR #102, opened 2018-10-29
Error notifications

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8047
https://issues.jboss.org/browse/AEROGEAR-7964

## Verification Steps
Verify that there are correct error messages shown in UI.

1. `./scripts/development.sh`
2. Try to create app with name that already exist
3. Open url for app name that does not exist
4. Try to kill backend with `lsof -i tcp:4000 | grep LISTEN | awk '{print $2}' | xargs kill`
5. restart MDC
6. `oc logout`

## Checklist:

### DIFF
--- a/pkg/web/buildConfigsHandler.go
+++ b/pkg/web/buildConfigsHandler.go
@@ -198,22 +198,27 @@ func newBuildConfigObject(data BuildConfigCreateRequest, namespace string) (*mob
 func (mbch *MobileBuildConfigsHandler) Create(c echo.Context) error {
 	reqData := new(BuildConfigCreateRequest)
 	if err := c.Bind(reqData); err != nil {
+		c.Logger().Errorf("error creating build config: %v", err)
 		return c.String(http.StatusBadRequest, err.Error())
 	}
 	if err := c.Validate(reqData); err != nil {
+		c.Logger().Errorf("error creating build config: %v", err)
 		return c.String(http.StatusBadRequest, err.Error())
 	}
 	config, secrets, err := newBuildConfigObject(*reqData, mbch.namespace)
 	if err != nil {
+		c.Logger().Errorf("error creating build config: %v", err)
 		return c.String(http.StatusInternalServerError, err.Error())
 	}
 	config, err = mbch.buildConfigsCRUDL.Create(config)
 	if err != nil {
+		c.Logger().Errorf("error creating build config: %v", err)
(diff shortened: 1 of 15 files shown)


pkg/web/buildConfigsHandler.go
```diff
@@ -198,23 +198,23 @@ func newBuildConfigObject(data BuildConfigCreateRequest, namespace string) (*mob
 func (mbch *MobileBuildConfigsHandler) Create(c echo.Context) error {
 	reqData := new(BuildConfigCreateRequest)
 	if err := c.Bind(reqData); err != nil {
-		return err
+		return c.String(http.StatusBadRequest, err.Error())
```
wei-lee: can we log the errors as well, or is it already logged?

jhellar: in the log there is just the http response code without the error message. Is it enough or we should also log the error message?

wei-lee: I think we need to error message.

ui/src/containers/ErrorMessages.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import { ToastNotificationList, ToastNotification } from 'patternfly-react';
+import _ from 'lodash-es';
+import { withRouter } from 'react-router-dom';
+import { dismiss, dismissAll } from '../actions/errors';
+import { wsError } from '../DataService';
+
+class ErrorMessages extends Component {
```
wei-lee: can we add a test for this component?

ui/src/reducers/resource.js
```diff
@@ -4,140 +4,161 @@ import {
...
+      errors = [...state.errors];
+      errorsToDismiss.forEach(e => {
+        index = errors.findIndex(err => err === e);
+        errors = [...errors.slice(0, index), ...errors.slice(index + 1)];
+      });
+      return {
+        ...state,
+        errors
+      };
+    case DISMISS_ALL_ERRORS:
```
wei-lee: is this actually used anywhere?

jhellar: it is called when route changes - https://github.com/aerogear/mobile-developer-console/pull/102/files#diff-0c044acc5817c9643bf3bb658b2d58c2R12

PR #106, opened 2018-11-02
Added minishift and oc-cluster-up install scripts

jhellar:
## Motivation

Currently for mac users we don't have recommendation for how to setup mobile.next with local OpenShift. The problem with minishift was with certificates installed on router by default. Because of this issue Android/iOS prevented communication with services running in OpenShift. This PR adds script that will set up minishift with correct certificates.

## Verification steps

1. `./scripts/minishift.sh`
2. deploy mdc and keycloak
3. create app and bind it with keycloak
4. verify that showcase app can communicate with keycloak

### DIFF
--- a/scripts/minishift.sh
+++ b/scripts/minishift.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")"
+BASEDIR="$(pwd)"
+
+minishift delete -f
+
+MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --openshift-version v3.11.0 \
+--extra-clusterup-flags "--enable=*,service-catalog,automation-service-broker,template-service-broker"
+
+oc login -u system:admin
+
+oc project default
+
+rm -rf /tmp/mini-certs
+mkdir /tmp/mini-certs
+cd /tmp/mini-certs
+
+oc get secret router-certs --template='{{index .data "tls.crt"}}' -n default  |  \
+base64 --decode | sed -e '1,/^-----END RSA PRIVATE KEY-----$/ d'  >localcluster.crt
+
+minishift ssh -- cat /var/lib/minishift/base/openshift-controller-manager/ca.crt >ca.crt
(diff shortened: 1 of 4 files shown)


scripts/minishift.sh
```diff
@@ -0,0 +1,61 @@
...
+
+cd "$(dirname "$0")"
+BASEDIR="$(pwd)"
+
+minishift delete -f
+
+MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --openshift-version v3.11.0 \
+--extra-clusterup-flags "--enable=*,service-catalog,automation-service-broker,template-service-broker"
+
+oc login -u system:admin
```
wei-lee: it needs to specify the target host here. Otherwise this will try to login to an existing oc session (which could point to a different cluster).

wei-lee:
@jhellar Do I need to use the latest version of minishift? Getting an error on start: https://gist.github.com/wei-lee/f6d43ef2736f73510ab6e5742bd9f154

scripts/minishift.sh
jhellar: it should not be needed, when `minishift start` completes, oc configuration should target this newly created cluster

wei-lee: ah, ok. The problem I have is that the minshift start command failed. So I think before this command, the script should fail if minishift failed to start?

jhellar:
@wei-lee not sure what causes the error, I have `minishift v1.25.0+90fb23e`, can you try to update to the latest minishift?

I've just updated this PR with few more changes.

scripts/minishift.sh
wei-lee: looks like it's ok, so don't worry about it.

jhellar: should be fixed now

wei-lee:
@jhellar verified. It is working now after I updating minishift.

wei-lee:
@jhellar another question come to my mind is if I restart minishift, what command should I run? Is it just run `minishift start`? Might be good to add that to the README

PR #107, opened 2018-11-07
[WIP] Load mobile client config to the UI

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8033
https://issues.jboss.org/browse/AEROGEAR-8089

## Checklist:

## Progress

- [] Finished task
- [ ] TODO


pkg/mobile/secrets.go
```diff
@@ -18,3 +20,15 @@ func NewSecretsCRUDL(client kubev1.CoreV1Interface) *SecretsCRUDLImpl {
...
 }
+
+func (crudl *SecretsCRUDLImpl) List(namespace string) (*v1.SecretList, error) {
+	listOpts := metav1.ListOptions{}
+	return crudl.secretsClient.Secrets(namespace).List(listOpts)
+}
+
+func (crudl *SecretsCRUDLImpl) Watch(namespace string) func() (watch.Interface, error) {
+	return func() (watch.Interface, error) {
+		watchOpts := metav1.ListOptions{}
```
wei-lee: we can probably use the `LabelSelector` in the `ListOptions` to only watch the secrets that we are interested in.

pkg/stub/handler.go
```diff
@@ -1,102 +1,119 @@
...
-	"github.com/operator-framework/operator-sdk/pkg/sdk"
-	log "github.com/sirupsen/logrus"
+	"github.com/labstack/gommon/log"
 	"k8s.io/api/core/v1"
 )
 
-func NewHandler(mobileAppRepo mobile.MobileClientRepo) sdk.Handler {
-	return &Handler{
-		mobileClientRepo: mobileAppRepo,
+func WatchSecrets(namespace string, secretsCRUDL mobile.SecretsCRUDL, mobileAppRepo mobile.MobileClientRepo) {
```
wei-lee: I think without too much effort, we can refactor this function a bit to improve the reusability: it can take in a `watchInterface`, and a listener function. Then this function can invoke the passed in `watchInteferace` to start watching, and invoke the listener function when it should be called.

pkg/stub/handler.go
```diff
@@ -1,102 +1,119 @@
...
+				continue
+			}
+			if willHandle {
+				continue
+			}
+			willHandle = true
+			timer := time.NewTimer(time.Second)
+			go func(timer *time.Timer) {
+				<-timer.C
+				willHandle = false
```
wei-lee: this probably should be set to `true` after the next line is finished execution? otherwise there is a chance the next line could be invoked in parallel? Also I am not sure if this need to executed in the background? why not just execute it inside the loop? This way we don't need to use the `willHandle` flag.

pkg/stub/handler.go
```diff
@@ -1,102 +1,119 @@
...
-
-		app, err := h.mobileClientRepo.ReadByName(clientId)
-		if err != nil {
-			log.Errorf("failed to read mobile client by name %v", clientId)
-			return err
+	}()
+	willHandle := false
+	for {
+		select {
+		case _, ok := <-events:
```
wei-lee: from my testing, this is getting triggered many times on server start. I think it is because the API returned all the events from the beginning. Maybe we should update the watch option so that it will only return the latest events?

pkg/stub/handler.go
jhellar: It is kind of a hack which makes sure, that in case there are e.g. 10 events at the same time (noticed that this happens e.g. when server starts) we will execute the handler function only once.

pkg/mobile/secrets.go
jhellar: ok, will add it

pkg/stub/handler.go
jhellar: makes sense, will refactor it

pkg/stub/handler.go
```diff
@@ -1,102 +1,119 @@
...
+	}
+	changed := false
+	for _, newService := range newServices {
+		serviceID := newService.Id
+		existing, _ := findService(existingServices, serviceID)
+		if existing == nil || existing.Version != newService.Version {
+			changed = true
+			break
 		}
+	}
```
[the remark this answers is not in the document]
jhellar: I'm not 100% sure, but I think that the current implementation handles this by first comparing number of new/existing services for an app. If they differ we will need to update services for that app. If they are same, the code is checking if every new service is also in existing services. If so (I hope that) this then implies that the new/existing services are the same and there is no need for an update.

jhellar:
@wei-lee thanks for review. I've pushed commit with fixes. Can you look again? The mobil-services.json is still just mock, only services in it are real data - I will send a separate PR for that later.

pkg/stub/handler.go
wei-lee: maybe try use the `ResourceVersion` field in the `ListOptions` and see if it can help?

pkg/stub/handler.go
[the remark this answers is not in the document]
wei-lee: yeah, gotcha.

pkg/stub/handler_test.go
```diff
@@ -86,206 +143,23 @@ func createSecret(labels map[string]string, data map[string][]byte, version stri
...
-				a, _ := mockAppRepo.ReadByName("test-app")
-				err := tc.ValidateClient(a)
-				if err != nil {
-					t.Fatalf("mobile client validation failed")
-				}
-			}
-		})
+	mockApps.Create(client)
+	go WatchSecrets("test-namespace", mockSecrets, mockApps)
+	time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
jhellar: should be fixed now

pkg/mobile/secrets.go
```diff
@@ -18,3 +20,17 @@ func NewSecretsCRUDL(client kubev1.CoreV1Interface) *SecretsCRUDLImpl {
...
+
+func (crudl *SecretsCRUDLImpl) List(namespace string) (*v1.SecretList, error) {
+	listOpts := metav1.ListOptions{}
+	return crudl.secretsClient.Secrets(namespace).List(listOpts)
+}
+
+func (crudl *SecretsCRUDLImpl) Watch(namespace string) func() (watch.Interface, error) {
+	return func() (watch.Interface, error) {
+		watchOpts := metav1.ListOptions{
+			LabelSelector: "mobile",
```
wei-lee: is this going to only return the secrets that have a `mobile` label?

jhellar: I hope so, haven't tested that though. will give it a try

pkg/stub/watch.go
```diff
@@ -0,0 +1,48 @@
...
+				}
+				events = watchInterface.ResultChan()
+				continue
+			}
+			if willHandle {
+				continue
+			}
+
+			// this makes sure that there will be only one call of handleChange in case there are more
+			// events notifying about change in the same time
```
wei-lee: right, now I understand.

In this case, would it be better if we keep the timestamp of when the last event was received, and if the time of the new event is too close to the previous one, we just ignore it? It's the same idea, but might be a bit easier to understand.

pkg/stub/secretsChangeHandler_test.go
```diff
@@ -0,0 +1,173 @@
...
+	app, _ := mockApps.ReadByName("test-app")
+	if len(app.Status.Services) != 1 {
+		t.Fatalf("mobile client validation failed")
+	}
+	mockSecrets2 := NewMockSecretsCRUDL2()
+	HandleSecretsChange("test-namespace", mockApps, mockSecrets2)
+	app, _ = mockApps.ReadByName("test-app")
+	if len(app.Status.Services) != 0 {
+		t.Fatalf("mobile client validation failed")
+	}
```
wei-lee: if it's not too much trouble, can we use the same pattern for the test cases as we did in other tests. It will make it easier to see cases we have covered, and adding new test cases.

pkg/mobile/secrets.go
jhellar: So this will watch only for changes in secrets with `mobile` label. But we then list all the secrets, so we still need to filter them.

pkg/stub/watch.go
jhellar: IMHO we would still need the goroutine to execute the handle function after the 1 second. if we execute it with the first event, the later changes (which will happen in that 1 second timeframe) might be ignored (not reflected in mobile-services.json files).

jhellar:
@wei-lee tests should be fixed now. can this be merged now?

PR #109, opened 2018-11-08
Adding model classes

ziccardi:
# Motivation
https://issues.jboss.org/browse/AEROGEAR-8002

## Description
Class has been created to resemble the JSON so that object receiving the JSON will continue to work even when receiving this object (but some edge case) to make the transition as smooth as possible.


ui/src/model/mobileapp.js
```diff
@@ -0,0 +1,79 @@
...
+  getSpec() {
+    return new MobileAppSpec(this.app);
+  }
+
+  toJSON() {
+    return { ...this.app };
+  }
+}
+
+export class MobileAppSpec {
```
wei-lee: We don't need to make those classes public. This is an internal class the the UI should not use.

ui/src/model/mobileapp.js
```diff
@@ -0,0 +1,79 @@
...
+  setType(newType) {
+    this.spec.clientType = newType;
+  }
+
+  setAppIdentifier(newAppIdentifier) {
+    this.spec.appIdentifier = newAppIdentifier;
+  }
+}
+
+export class MobileAppMetadata {
```
wei-lee: same for this one

wei-lee:
@wojta maybe there is. But the main purpose here is to hide the complexity of the `MobileApp` model on the backend from the frontend. If the frontend dealing with JSON objects directly, then they are forced to know what the internals of the `MobileApp` model, and make it hard to refactoring, add/remove fields in the future. But if we have a simple model for it on the frontend, we will be able to abstract the backend model away.

aliok:
@ziccardi I see some errors in the Chrome console when I open the UI.
Also, I can't create clients

aliok:
The problems I mentioned above were only in my side.

@ziccardi @wei-lee  can we merge this PR please?

ui/src/model/mobileapp/index.js
```diff
@@ -0,0 +1 @@
+export { MobileApp } from './mobileapp';
```
wei-lee: do we actually need this file?

ui/src/components/overview/MobileClientCardView.js
```diff
@@ -142,7 +143,7 @@ class MobileClientCardView extends Component {
             />
           </Filter>
           <div className="form-group">
-            <CreateClient platforms={SUPPORTED_PLATFORMS} />
+            <CreateClient platforms={SUPPORTED_PLATFORMS} item={new MobileApp().toJSON()} />
```
wei-lee: maybe just have a default value in `CreateClient` component, rather than pass it in?

ziccardi: That's correct, however the `CreateClient` is going to be fully rewritten (no `platform selection`, no `appIdentifier`, etc) so I don't think it worths fixing this.

wei-lee: ok

aliok: that's right. I already removed the supported platforms param in my PR, which is based on this PR

PR #110, opened 2018-11-08
Setting default client id for binding properties

secondsun:
## Motivation
During the review John asked if the client ID could be set by default.  Now it is.

## What
The default client id is set when MobileServiceView is building the unbound service row objects.

## Why
Usability.

## How
The default client id is set when MobileServiceView is building the unbound service row objects.  The function is wrapped in a try catch block that does minimal logging.  Worst case scenario is the user has to input this value themself.

## Verification Steps
When you try to bind a mobile service, you should see on the second page of the wizrd that client id is now prefilled.

### DIFF
--- a/ui/src/components/mobileservices/MobileServiceView.js
+++ b/ui/src/components/mobileservices/MobileServiceView.js
@@ -9,6 +9,7 @@ class MobileServiceView extends Component {
     super(props);
     this.boundServiceRows = this.boundServiceRows.bind(this);
     this.unboundServiceRows = this.unboundServiceRows.bind(this);
+    this.setDefaultBindingProperties= this.setDefaultBindingProperties.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


ui/src/components/mobileservices/UnboundServiceRow.js
```diff
@@ -13,6 +13,7 @@ class UnboundServiceRow extends Component {
     }
 
     this.service = props.service;
+    console.log(this.service)
```
jhellar: this should be removed

ui/src/components/mobileservices/MobileServiceView.js
```diff
@@ -33,12 +34,23 @@ class MobileServiceView extends Component {
...
         rows.push(<UnboundServiceRow key={service.serviceId} service={service} />);
       });
     }
     return rows;
   }
 
+  setDefaultBindingProperties(service) {
+    try {
+      if (service.bindingSchema.properties["CLIENT_ID"]) {
+        service.bindingSchema.properties["CLIENT_ID"].default = this.props.appName;
```
wei-lee: is this field still showing up in the UI? Should we just hide it given that user should not need to change it anyway?

secondsun: The answer to this question is long and complicated.  

The field is required by the binding schema provided in the APB.  Even though we can prefill the information, I don't feel right about hiding what we are doing from the user.  We might instead remove the client id property from the binding schema and set that as part of creating the binding on the server side, but I don't know what knock on effects that will cause.

I am fine with making the field not editable.

wei-lee: if it's not editable, and the fact that the app name is showing up on the screen, I do think there is not too much value to show this field. I still think we should make it not visible. It will keep the binging process clean.

secondsun: Oh new info, we can't remove the field because it is used by openshift's binding UI.
image

secondsun: image

secondsun: My "I don't think it is a good idea" has just been upgraded to "doing this is a terrible idea"

PR #114, opened 2018-11-12
Load mobile client config to the UI

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8033

## What
Remove mock data for mobile-services.json. Update clusterName, namespace and clientID with real data.

### DIFF
--- a/cmd/api-server/main.go
+++ b/cmd/api-server/main.go
@@ -91,7 +91,7 @@ func main() {
 	}
 
 	{
-		mobileClientsHandler := web.NewMobileClientsHandler(mobileClientsRepo, namespace)
+		mobileClientsHandler := web.NewMobileClientsHandler(mobileClientsRepo, namespace, cfg.Host)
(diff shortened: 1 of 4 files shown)


wei-lee:
👀

ui/src/containers/ConfigurationView.js
```diff
@@ -0,0 +1,84 @@
...
+  const { status = {} } = props.app;
+  const { services = [] } = status;
+  const { metadata = {} } = props.app;
+  const config = {
+    version: 1,
+    clusterName: props.openshiftMasterUrl,
+    namespace: metadata.namespace || '',
+    clientId: metadata.name || '',
+    services: []
+  };
```
wei-lee: The information is already available on the backend [here](https://github.com/aerogear/mobile-developer-console/blob/master/pkg/web/mobileClientsHandler.go#L74). Can we not update the backend to return the correct value, rather than override them on the frontend here?

jhellar:
@wei-lee thanks for pointing that out, should be fixed now.

PR #116, opened 2018-11-13
Fixing eslint

ziccardi:
# Motivation

https://issues.jboss.org/browse/AEROGEAR-7960

# Description

- Fixed `eslint` errors
- Added `eslint` to the build to avoid linter errors to accumulate again


ziccardi:
@wei-lee just done a lot of `autofix`, manually reorganized imports (absolute import first) and guarded some `for loop`. I tested the app and looks like it works. Going to merge it now to avoid conflicts.

PR #117, opened 2018-11-13
 ✨ [WIP] show the status of a bind action

wei-lee:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8090

## How

Show the status of a bind action (like "In Progress", or "Failed"). If a bind is successful, it will show up as a "Bound service". Otherwise if the bind is failed, user should have the option to try again.

## Progress

- [x] Add the mobile service models
- [ ] Update UI to show the bind status

@secondsun this is still in progress, didn't get the chance to complete all tasks, but I will continue tomorrow. Want to let you know that I have created some model classes for the mobile services.


ui/src/models/mobileService.js
```diff
@@ -0,0 +1,96 @@
...
+
+  getFormDefinition() {
+    return this.servicePlan.spec.externalMetadata.schemas.service_binding.create.openshift_form_definition;
+  }
+
+  getServiceClassExternalName() {
+    return this.serviceClass.spec.externalMetadata.serviceName;
+  }
+
+  getCurrentBindStatus() {
```
secondsun: Oh hey, so I'm guessing this will be replacing/extending the boolean IsBound (from the go code) with an actual status enum of some kind?

wei-lee: yes, my plan is to get the `conditions` field and see what is the status of the binding. It should be one of these 2 status: `inprogress` or `failed`. If the binding is completed successfully, it will be a `BoundMobileService`.

wei-lee:
screen shot 2018-11-14 at 12 56 01

wei-lee:
@secondsun The UI is updated now. I will add some unit tests after lunch.

wei-lee:
@secondsun mind reviewing this? I will add tests as part the other JIRA.

secondsun:
@wei-lee looking now

wei-lee:
@secondsun @ziccardi I refactored the models for the UI a bit, could you both please take a look. Thanks.

ui/src/models/mobileapps/mobileapp.js
```diff
@@ -43,18 +38,11 @@ export class MobileApp {
   }
 
   setProperty(propertyName, propertyValue) {
-    switch (propertyName) {
-      default:
-        this.getSpec().set(propertyName, propertyValue);
-        break;
-    }
+    this.getSpec().set(propertyName, propertyValue);
```
ziccardi: This way we don't abstract the JSON structure anymore. I mean: when we have to set the name of the application, I would expect to be able to do `model.setProperty('newName')`, while with this code I have to do `model.setProperty('spec.name', 'newName')` and, in my opinion, we lose the usefulness of the model classes.
The idea that was behind the `switch/case` was that we could add there all the property whose path we want to abstract.

wei-lee: I am not sure I follow you here. The method is the same as before, just the implementation a bit different

ziccardi: I put the comment here in the `setProperty`, but it applies to the `getProperty` as well. Suppose you want to get the version of the app. With the current implementation, we can't use the `getPropery` (nor the `setProperty`), but we have to get the `Status` object and get the version from there.
The idea behind the `switch/case` was that you could do `getProperty('version')` and we just had to add a 
```js
case 'version':
  return this.getStatus().getVersion();
```

wei-lee: ok, if this is the case, then should we not move all the methods in `AppStatus` and `AppSpec` to `MobileApp`, and not exposing the `status` or `spec` fields at all?

ziccardi: That's why the `get/setProperty` method is there. And yes, spec and status were not expected to be exposed (at least, that was the idea).

wei-lee: ok, understood. Given that we are only support get/set properties on the spec object, I am going to leave as it is . But I will add a comment to the code to explain that this can be extended to support set/get more properties on other objects.

ziccardi:
👀

secondsun:
@wei-lee If the binding is in progress, should we hide the bind app button?

wei-lee:
> @wei-lee If the binding is in progress, should we hide the bind app button?

@secondsun yes, the button should be hidden when there is a bind in progress. You can see it in the screenshot. Are you still seeing the button?

secondsun:
@wei-lee Oh!  I must have had an old window open, it is gone now...

PR #118, opened 2018-11-14
oc-cluster-up fix

jhellar:
## Motivation

`oc-cluster-up` script was not properly tested when added. This PR should fix this script. Tested here https://mobile-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/mobile-next-test/22/console

## Changes

- cluster ip can now be provided to the script (needed for ci)
- added `no-proxy` flag for `oc cluster up` (needed in ci environment)
- additional services (e.g. ansible-service-broker) installed separately from `oc cluster up` - this enables rerunning the script
- make sure temporary files are editable with `post-install` script (were blocked by se-linux)
- fix `setup-router-certs` script for oc cluster up

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,17 @@ You can follow the docs [here](https://github.com/aerogear/mobile-core)
 
 :penguin: Linux
 
+You may need to configure your firewall first:
+
+```
+sudo firewall-cmd --permanent --add-port=8443/tcp
+sudo firewall-cmd --permanent --add-port=8053/tcp
+sudo firewall-cmd --permanent --add-port=53/udp
+sudo firewall-cmd --permanent --add-port=443/tcp
+sudo firewall-cmd --permanent --add-port=80/tcp
+sudo firewall-cmd --reload
+```
+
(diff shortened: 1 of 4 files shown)


aliok:
@jhellar  thanks for the good work.

I would say it is in a much better state than before, but I still get some errors...

See this screenshot:
screenshot from 2018-11-14 12-03-48

This happens after I enter the creds in the console login form.

Haven't seen this before.

But, things work with `oc` client. I wasn't able to login with `system:admin` before and now it works.

Here's a snippet I got from @ziccardi (modified a bit):
```
oc cluster down
rm -rf openshift.local.clusterup
oc cluster up
oc whoami
oc login -u system:admin && oc adm policy add-cluster-role-to-user access-asb-role developer && oc adm policy add-cluster-role-to-user cluster-admin developer && oc login -u developer
oc create -f deploy/crd.yaml
oc new-project mdc
export NAMESPACE=mdc
export KUBERNETES_CONFIG=$HOME/.kube/config
oc process -f mobile-developer-console.template.yaml | oc create -f -
./scripts/development.sh
```

This script has a different purpose and less requirements of course (doesn't care about the cluster ip nor the certs).

PR #120, opened 2018-11-14
AEROGEAR-8100 - App platforms no more

aliok:
AEROGEAR-8100 

* Deleted app platforms from everywhere except build related things.
* Didn't want to touch build related things because we don't know if we are going to support them in the near future.
* Also didn't want to remove app identifier from anywhere. I will do that in a separate PR.
* Platform/clientType is deleted from the CRD definition as well. I didn't make it optional. IMO, making it optional would just create confusion.

Verification:
* No icon in the mobile app cards in the overview page
* No platform selection during new app creation and in edit
* Backend adapted: create and edit works


ziccardi:
👀

ui/src/components/overview/MobileClientCardView.js
```diff
@@ -47,7 +39,7 @@ class MobileClientCardView extends Component {
       <EmptyState>
         <EmptyStateTitle>{this.emptyStateMessage.noAppsCreated}</EmptyStateTitle>
         <EmptyStateAction>
-          <CreateClient createButtonSize="large" platforms={SUPPORTED_PLATFORMS} item={new MobileApp().toJSON()} />
+          <CreateClient createButtonSize="large" item={new MobileApp().toJSON()} />
```
ziccardi: As suggested in another PR by @wei-lee, change the `item` to be a defaulted to `new MobileApp().toJSON()` instead of passing it here

ui/src/components/overview/MobileClientCardView.js
```diff
@@ -143,7 +135,7 @@ class MobileClientCardView extends Component {
             />
           </Filter>
           <div className="form-group">
-            <CreateClient platforms={SUPPORTED_PLATFORMS} item={new MobileApp().toJSON()} />
+            <CreateClient item={new MobileApp().toJSON()} />
```
ziccardi: Look previous comment

ui/src/components/overview/MobileClientCardView.js
aliok: Let me address that later.

aliok: This commit fixes this: https://github.com/aerogear/mobile-developer-console/pull/124/commits/35e9d0b4f4b4b56775bdf3f25df1de7d8e54bd82

PR #122, opened 2018-11-14
Aerogear 8086 update binding forms

secondsun:
## Motivation
See https://issues.jboss.org/browse/AEROGEAR-8086

## What
This PR adds binding form validation and updates the forms to make the client id not editable, as well as make the passphrase fields be passwords with a match and the certificate be a text area.

## Why
This was bumped from the previous sprint

## How
A validaiton method was added to BindingPanel.js and the ObjectTemplate in bindingpanelutils.js was updated to perform more custom rendering

## Verification Steps
Create each of the bindings for Metrics, Push, Sync, and Keycloak.  Confirm they bind correctly.

### DIFF
--- a/ui/src/components/mobileservices/BindingPanel.js
+++ b/ui/src/components/mobileservices/BindingPanel.js
@@ -20,15 +20,15 @@ class BindingPanel extends Component {
   }
 
   onNextButtonClick() {
-    const {  activeStepIndex } = this.state;
+    const { activeStepIndex } = this.state;
     if (activeStepIndex === 1) {
-      this.form.submit()
-      return false;//swallow the event, see validate function
-    } else {
-      this.setState({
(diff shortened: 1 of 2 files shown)


ui/src/components/mobileservices/BindingPanel.js
```diff
@@ -65,9 +71,12 @@ class BindingPanel extends Component {
     
           return <Form schema={this.state.schema}
```
wojta: I should have used this in Create Build Config form instead of implementing it by myself, it could saved me much time.

secondsun:
@wei-lee Was a small css error.  I just removed the * since the only required fields are either the client ID or drop downs we don't let not have a value.  Updated and rebased

wei-lee:
@wojta the issue you are talking about is fixed in #117

PR #125, opened 2018-11-15
Services JS SDK docs generated based on bound services on the Configuration tab

wojta:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8032
https://issues.jboss.org/browse/AEROGEAR-8035
https://issues.jboss.org/browse/AEROGEAR-8036

## What
Information about AeroGear services configuration is generated based on bound services in the Configuration tab.

## Why
To make it easy for developers to set up their project.

## How
Mobile app information contains information about types of bound services. There is documentation stored in `ui/src/components/configuration/sdk-config-docs`, planned to support more frameworks in future.

## Verification Steps
 
1. Go to some mobile app
2. On the *Configruation* tab, you should see generated documentation
3. Try to create new service binding or delete existing one
4. Documentation should update based on the bound services

## Checklist:

## Progress

- [x] Implementation for Cordova docs
- [x] Add other supported frameworks?
- [x] Update view based on flats
- [x] SDK services docs links
- [x] SDK services docs links with AEROGEAR_DOCS_VERSION env. variable
- [x] Add tests for some components used

## Additional Notes
Cordova is now hardcoded, it needs to be updated with info about what frameworks will be supported.
image


wei-lee:
👀

ui/src/components/configuration/sdk-config-docs/cordova.js
```diff
@@ -0,0 +1,107 @@
...
+        };
+        
+        init(aerogearConfig);`
+        ]
+      ]
+    },
+    {
+      introduction:
+        'Copy `mobile-services.json` content and copy it to the location `src/mobile-services.json` in your application project.'
+    }
```
wei-lee: I think the instructions here are not accurate. Especially around when to copy the `mobile-services.json` file and how to init. The right order should be (after installing the core package):

1. Copy `mobile-services.json` file to the app's source folder.
2. To init, load the content from `mobile-services.json` file, and pass it to init:
    ```
    import { init } from "@aerogear/app";
    const appConfig = require("./mobile-services.json");
    init(appConfig);
    ```

wei-lee:
@wojta ok, let's wait for it to be fixed in upstream. It is not a blocker, so let's merge the PR then!

PR #127, opened 2018-11-19
When delete a service binding, the view is not updated

ziccardi:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8128

## Progress

- [x] Move service bindings reducers to its own module
- [x] Change the reducer to put only binding data into the binding data store

### Binding
1. Click on an existing application to see its details
2. Click on mobile services
3. Bind a new service
5. Verify the the dialog closes and the service is put in 'Binding in progress'

### Unbinding
1. Click on an existing application to see its details
2. Click on mobile services
3. If the application has no services bound, bind a new one
4. Delete one bound service
5. Verify the the dialog closes and the service is put in 'Unbinding in progress'

**NOTE:** Currently the service will remain to the `in progress` state until you refresh. @secondsun is working on a Websocket implementation that should receive events from the server

### DIFF
--- a/ui/src/actions/serviceBinding.js
+++ b/ui/src/actions/serviceBinding.js
@@ -25,16 +25,14 @@ export const createBinding = (
   fetchAction(
     [SERVICE_BINDING_CREATE_REQUEST, SERVICE_BINDING_CREATE_SUCCESS, SERVICE_BINDING_CREATE_FAILURE],
     async () =>
-      DataService.mobileApp(formData.CLIENT_ID).then(mobileApp =>
-        DataService.createBinding(
-          mobileClientName,
-          serviceInstanceName,
-          credentialSecretName,
-          parametersSecretName,
-          serviceClassExternalName,
-          formData
-        ).then(binding => ({ mobileApp, binding, serviceInstanceName }))
-      )
+      DataService.createBinding(
+        mobileClientName,
+        serviceInstanceName,
+        credentialSecretName,
+        parametersSecretName,
+        serviceClassExternalName,
+        formData
+      ).then(binding => ({ binding, serviceInstanceName }))
   )();
 
 export const SERVICE_BINDING_DELETE_REQUEST = 'SERVICE_BINDING_DELETE_REQUEST';
--- a/ui/src/actions/serviceBinding.js
+++ b/ui/src/actions/serviceBinding.js
@@ -25,13 +25,15 @@ export const createBinding = (
   fetchAction(
     [SERVICE_BINDING_CREATE_REQUEST, SERVICE_BINDING_CREATE_SUCCESS, SERVICE_BINDING_CREATE_FAILURE],
     async () =>
-      DataService.createBinding(
-        mobileClientName,
-        serviceInstanceName,
-        credentialSecretName,
-        parametersSecretName,
-        serviceClassExternalName,
-        formData
+      DataService.mobileApp(formData.CLIENT_ID).then(mobileApp =>
+        DataService.createBinding(
+          mobileClientName,
+          serviceInstanceName,
+          credentialSecretName,
+          parametersSecretName,
+          serviceClassExternalName,
+          formData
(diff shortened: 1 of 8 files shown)


wei-lee:
👀

ui/src/reducers/serviceBindings.js
```diff
@@ -9,18 +9,91 @@ import {
...
+      return {
+        ...state,
+        isCreating: true
+      };
+    case SERVICE_BINDING_CREATE_SUCCESS:
+      return {
+        ...state,
+        isCreating: false,
+        errors: getErrors(null, 'create', state.errors)
+      };
```
wei-lee: when a bind is created, even though it is still being progress on the backend, the UI state should be updated: there is now a new binding, and its status is ""inprogress", the UI state should reflect this change.

ui/src/reducers/serviceBindings.js
```diff
@@ -9,18 +9,91 @@ import {
...
+      return {
+        ...state,
+        isDeleting: true
+      };
+    case SERVICE_BINDING_DELETE_SUCCESS: {
+      return {
+        ...state,
+        isDeleting: false,
+        errors: getErrors(null, 'delete', state.errors)
+      };
```
wei-lee: similar to my comment above, at this point, the UI should be updated to reflect the change that a binding is being removed.

ziccardi:
@wei-lee I addressed the issues you described in the comments. Mind giving another look?
@secondsun : I believe this should complement your PR

ui/src/actions/serviceBinding.js
```diff
@@ -25,13 +25,15 @@ export const createBinding = (
...
     [SERVICE_BINDING_CREATE_REQUEST, SERVICE_BINDING_CREATE_SUCCESS, SERVICE_BINDING_CREATE_FAILURE],
     async () =>
-      DataService.createBinding(
-        mobileClientName,
-        serviceInstanceName,
-        credentialSecretName,
-        parametersSecretName,
-        serviceClassExternalName,
-        formData
+      DataService.mobileApp(formData.CLIENT_ID).then(mobileApp =>
```
wei-lee: why do we need to fetch the app here first?

ui/src/reducers/serviceBindings.js
```diff
@@ -9,18 +11,129 @@ import {
...
+    case SERVICE_BINDING_CREATE_REQUEST:
+      return {
+        ...state,
+        isCreating: true
+      };
+    case SERVICE_BINDING_CREATE_SUCCESS: {
+      const { serviceInstanceName } = action.result;
+      const index = findIndex(
+        state.unboundServices,
+        binding => binding.serviceInstance.metadata.data.name === serviceInstanceName
```
wei-lee: there is a `getServiceInstanceName` available on the `UnboundMobileService` model, use that instead.

ui/src/reducers/serviceBindings.js
```diff
@@ -9,18 +11,129 @@ import {
...
+    }
+    return errors;
+  }
+  if (index >= 0) {
+    return [...errors.slice(0, index), { error, type }, ...errors.slice(index + 1)];
+  }
+  return [{ error, type }, ...errors];
+};
+
+const configureInProgress = (serviceBinding, currentOperation) => ({
```
wei-lee: should we move this to the model as a method on the model instead?

ui/src/reducers/serviceBindings.js
```diff
@@ -9,18 +11,129 @@ import {
...
+  ...serviceBinding,
+  isBound: false,
+  serviceBinding: { status: { currentOperation, conditions: [{ type: 'Ready', status: 'False' }] } }
 });
 
-export default serviceBindings;
+const serviceBindingsReducer = (state = defaultState, action) => {
```
wei-lee: can we also add a unit test for this? it should be very easy to write.

ziccardi:
@wei-lee : added the tests and addressed the comments. Mind giving another look?

PR #128, opened 2018-11-19
Installer fix - waiting for asb deployment

jhellar:
## Motivation

https://trello.com/c/zQBCWwKI/25-1-replace-waiting-for-asb-to-spin-up-with-canceling-deployment-in-installer

- canceling of deployment is not supported in oc 3.11
- current implementation checks number of running pods which does not work properly every time
- replaced with deployment status check

https://trello.com/c/Dbyr6z40/42-1-get-name-of-asb-project-in-installer-automatically

Also added condition that checks if ASB was already configured, if so the config is skipped.

### DIFF
--- a/scripts/post_install.sh
+++ b/scripts/post_install.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
 cd "$(dirname "$0")" || exit 1
+ASB_PROJECT_NAME=$(oc get projects | grep -e ansible-service-broker -e automation-service-broker | awk '{print $1}')
 if [ -z "$ASB_PROJECT_NAME" ]
 then
-    echo "Please specify OpenShift ASB project before running this script. E.g.:"
-    echo "export ASB_PROJECT_NAME='openshift-automation-service-broker'"
+    echo "Ansible service broker project not found."
     exit 1
 fi
 oc project "$ASB_PROJECT_NAME" || exit 1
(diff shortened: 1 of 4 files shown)


scripts/post_install.sh
```diff
@@ -16,24 +16,25 @@ UPDATE_INSTRUCTIONS="yml/update_instructions.yaml"
...
-# Replace double quotes with single quotes
-sed -i.bak -e "s/\"/'/g" $ASB_TEMPLATE_FILENAME
-oc patch configmap broker-config -p "{\"data\":{\"broker-config\": \"$(awk '{printf "%s\\n", $0}' ${ASB_TEMPLATE_FILENAME})\"}}" -n "$ASB_PROJECT_NAME"
-oc rollout latest "$ASB_DC_NAME" -n "$ASB_PROJECT_NAME"
\ No newline at end of file
+if ! $yq r $ASB_TEMPLATE_FILENAME registry.*.org | grep aerogearcatalog &>/dev/null ; then
+    while oc rollout history dc/"$ASB_DC_NAME" -n "$ASB_PROJECT_NAME" | grep Running &>/dev/null
+    do
+        echo "Waiting for ASB to spin up" && sleep 5
+    done
```
wei-lee: why do we need to wait for the ASB to be up & running, before we can make the config change?

jhellar: after the config map is changed we need to redeploy the ASB, but this can not be done if deployment is running

PR #133, opened 2018-11-21
fix: integration test (clientType deprecation)

psturc:
## Why
Integration tests were failing because of deprecated `clientType` property

## Verification Steps


wei-lee:
@psturc is this one ready to merge? if so can you rebase & merge it.

PR #136, opened 2018-11-21
[SPIKE] Filter display group by

secondsun:
@wei-lee @ziccardi , Wei and I were discussing what you and I were discussing Massimo and I wanted to try to find a more generic way to do the filtering and I think I found it.

This PR works in conjuction with editing the cluster service plan for the UPS APB to add a "filterDisplayGroupBy":"CLIENT_TYPE" field to the mobileclient_bind_parameters_data if the apb.  (SEE https://github.com/aerogearcatalog/unifiedpush-apb/blob/master/apb.yml#L32 where it would go).

Anyway, I know last we spoke MZ I was against this.  Wei convinced me that I was still against this, but we needed it ;)  Here is my attempt to implement this without adding a lot of custom code.

If we adopt this PR we will need to update the apb.  You can update your own clusterserviceplan using `oc edit clusterserviceplan` as a cluster admin (ie system:admin)

## Demo
https://youtu.be/zByRFTe7FTI

### DIFF
--- a/ui/src/components/mobileservices/bindingPanelUtils.js
+++ b/ui/src/components/mobileservices/bindingPanelUtils.js
@@ -6,6 +6,7 @@ export const OpenShiftObjectTemplate = function({
   uiSchema,
   schema,
   properties,
+  formData,
   idSchema,
   title,
   description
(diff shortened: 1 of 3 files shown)


ziccardi:
@secondsun , adding a new field to the schema was exactly the work I was doing yesterday for the issue we discussed in the evening (https://issues.jboss.org/browse/AEROGEAR-8133).
Since you didn't like it, I was going to discuss this with @wei-lee  today.
However, you discussed AEROGEAR-8133 with @wei-lee yesterday evening while I was off (I received your mail at 9 pm) and you even implemented it: it doesn't make any sense that I continue with my coding.
I'm going to assign the issue to you and thrash the work I was doing.

ziccardi:
@secondsun Implemented here: https://github.com/aerogear/mobile-developer-console/pull/139

PR #139, opened 2018-11-22
Improved binding UI

ziccardi:


ziccardi:
@wei-lee Currently the validators only checks if the value is there or not.
It will be easy to add more fine graned checks into each validator (for example, to check that the provided P12 is really a base64 encoded P12 and that the password is correct)

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
```diff
@@ -0,0 +1,68 @@
...
+    this.detectPlatform = customProperties.platformDetector;
+    this.errorsKey = customProperties.errorField;
+    this.platform = platform;
+  }
+
+  addError(errors, key, message) {
+    errors[this.errorsKey || key].addError(message);
+  }
+
+  preFieldValidation(key, value) {
```
wei-lee: this one isn't used anywhere. Do we need it? When the `preFieldValidation` should be used? What's the difference between the preValidate/validateField/postValidate fields?

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        error = this.validateField(key, value);
+        if (error) {
+          hasErrors = true;
+          this.addError(errors, key, error);
+          continue;
+        }
+
+        error = this.postFieldValidation(key, value);
```
wei-lee: given that all the validators are called in a single loop, and the input and return values are exactly the same, I am not sure if there is much value to separate the validation methods to 3 different ones. It will also cause confusion on when to use which?

ui/src/components/mobileservices/bindingPanelUtils.js
```diff
@@ -123,7 +126,17 @@ function getFieldSet(field, properties, idSchema, schema) {
         return property.content;
     }
   });
-
+  if (uiSchema.form.filterDisplayGroupBy) {
+    if (field.title.toLowerCase().localeCompare(formData[uiSchema.form.filterDisplayGroupBy].toLowerCase()) === 0) {
```
wei-lee: `formData[uiSchema.form.filterDisplayGroupBy]` is possible to be undefined.

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
ziccardi: The execution order is:

1. preFieldValidation
2. validateField
3. postFieldValidation

both `preFieldValidation` and `postFieldValidation` are optional. The use of them is to perform some operation *before* the validation (`preFieldValidation`) or *after* the validation. If any of the method fails, the flow is interrupted.

Currently, only the `postFieldValidation` is used: if the `validateField` for the password field passes, then the password is matched to the `password confirmation`.

Since the confirmation field is a field that we add in the UI, it is not part of the `errors` object and custom code must be executed to get its value and validate it: the `postFieldValidation` helps in this case (see [here](https://github.com/aerogear/mobile-developer-console/blob/caabe8e32372a6166a770e04575aaf6e58a92d5d/ui/src/components/mobileservices/BindingPanel.js#L144))

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
ziccardi: The `preFieldValidation` and `postFieldValidation` are custom function that can be passed to the validators to perform custom `pre` or `post` validations.

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
wei-lee: I can understand, but personally I think it is a bit over-complicate: effectively they can all processed as part of validate, so I just don't get the benefit of splitting them. 

Anyway, given that `preFieldValidation` is not used anywhere, it's best to remove it for now. We can add it back later if we really need it.

Can we also add some comments in here to explain the difference.

ui/src/components/mobileservices/validators/AbstractBindingValidator.js
wei-lee: yes, but why custom? WE already have the concrete class to validate for android and ios, why split the validation logic in 2 different places?

ziccardi: To keep the interface consistent even in case we will have to add other validators.
Currently all the validations requires only the `formData` object, while only the password confirmation requires the `formData`, `form` and the `document` (see [here](https://github.com/ziccardi/mobile-developer-console/blob/caabe8e32372a6166a770e04575aaf6e58a92d5d/ui/src/components/mobileservices/BindingPanel.js#L146-L147) ). This way we can have a coherent interface even for future validators requiring other custom validations.

PR #140, opened 2018-11-23
UX fixes

wojta:
## Motivation
Some UX fixes based on https://issues.jboss.org/browse/AEROGEAR-8136

## Why
Improving user experience

## Checklist:

## Progress
General:

- [x] Clicking on "Mobile Developer Console" on top left should go to overview screen (similar to OKD console)
- [x] Documentation link in the question mark in top right should open in new window

Overview page:

- [x] Disable autocomplete in mobile client input UPDATE: not sure about this. OKD console has autocomplete not disabled in create project dialog
- [x] App filter in the overview page requires pressing enter. I would expect it to filter on the fly when typing.
- [x] Alignment of "Mobile Apps > myapp", "myapp" and the tabs below are not right. ...this is in the mobile client detail view.
Strange popups/tooltips in Configuration tab when you move mouse over to services displayed there.
- [x] Code block for mobile-services.json shows a gray bar on the left, which made me think it is disabled. Blue would be nice.
- [x] Create and Cancel buttons are in the wrong order in "Create build config" windows
Should display a message like "There are no bound services" when there are not. Same thing for unbound services.
- [x] Show "created 10 minutes ago" next to client name in the client detail view? Similar to OKD console.
Options


wei-lee:
👀

ui/src/components/common/NavHeader.js
```diff
@@ -1,14 +1,21 @@
 import React from 'react';
 import { Masthead, MenuItem, Icon } from 'patternfly-react';
-
+import { withRouter } from 'react-router-dom';
 import './NavHeader.css';
 
-const NavHeader = ({ title, user, helpDropdownItems, userDropdownItems }) => (
-  <Masthead title={title} navToggle={false}>
+export const _NavHeader = ({ title, user, helpDropdownItems, userDropdownItems, titleHref, history }) => (
+  <Masthead title={title} navToggle={false} onTitleClick={() => history.push(titleHref)}>
```
wei-lee: why not just use the "href" prop for `Masthead`?

ui/src/components/mobileservices/MobileServiceView.js
```diff
@@ -25,28 +27,29 @@ class MobileServiceView extends Component {
...
-    }
-
-    return rows;
+    return (
+      <React.Fragment>
+        <h2 key="bound-services">Bound Services</h2>
+        {this.props.boundServices && this.props.boundServices.length > 0 ? (
+          this.props.boundServices.map(service => <BoundServiceRow key={service.getId()} service={service} />)
+        ) : (
+          <h3>There are no bound services.</h3>
```
wei-lee: can we try `<EmptyState>` here and see what's it look like?

aliok:
I only have this suggestion:

*  Documentation link in the question mark in top right should open in new window: how about adding a https://fontawesome.com/icons/external-link-alt?style=solid icon to denote it opens in new page?


Great work @wojta !

I was testing "Show "created 10 minutes ago" next to client name in the client detail view? Similar to OKD console" and it autoupdates the text to 1 mins ago, 2 mins ago etc. when you just stay in the page.

I am super impressed.

ui/src/components/common/NavHeader.js
wojta: good catch, I was looking for it and missed it.

ui/src/components/mobileservices/MobileServiceView.js
wojta: image

PR #142, opened 2018-11-27
 🎨 make sure the data associated with an app is removed when the app is

wei-lee:
deleted

## Motivation
https://issues.jboss.org/browse/AEROGEAR-8141

## What

Remove all the data associated with an app when it's been deleted.

## Why

see jira

## How

Add label to bindings, build configs and secrets that are created for an app. When the app is delete, use the label to find the data that should be removed with the app

## Verification Steps

1. Provision some services and bind an app to the services.
2. Make sure you can see the bindings in the OpenShift web console
3. Remove the app
4. You should see the bindings are removed as well in OpenShift web console.

## Checklist:

### DIFF
--- a/pkg/mobile/bindableServices.go
+++ b/pkg/mobile/bindableServices.go
@@ -108,7 +108,10 @@ func (bms *BindableMobileServiceCRUDLImpl) DeleteAppData(mobileClient *v1alpha1.
 	}
 
 	for _, binding := range bindings.Items {
-		bms.Delete(mobileClient.GetNamespace(), binding.GetName())
+		err := bms.Delete(mobileClient.GetNamespace(), binding.GetName())
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/pkg/mobile/bindableServices.go
+++ b/pkg/mobile/bindableServices.go
@@ -2,40 +2,81 @@ package mobile
 
 import (
 	"encoding/json"
+	"fmt"
+	"github.com/aerogear/mobile-developer-console/pkg/apis/aerogear/v1alpha1"
 	"strings"
 
-	v1beta1 "github.com/kubernetes-incubator/service-catalog/pkg/apis/servicecatalog/v1beta1"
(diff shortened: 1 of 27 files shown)


wei-lee:
@secondsun I was thinking about it, but I didn't use that approach for a few reasons:

1. For bindings, there are actually 2 owners for each binding: the app, and the serviceInstance. I don't know what is the correct way to setup the owns in this case, and there isn't enough documents out there about the how to setup owners in this case. Also by looking at how delete service instance works in OpenShift Web console, they didn't use the owner field at all: they just remove bindings first, then remove the service instance. So I follow the same approach here: remove the bindings manually, and then remove the app.
2. There isn't too much differences code wise: I still have to update each object that's been created to add the owner field, the same as adding the labels. while it does save some effort so that we don't need to write the code to actually remove the associated objects, there are a few downsides:
* it's hard to know when the orphaned objects will be deleted: it will depend on how often the GC runs in k8s
* for someone who is not familiar with how k8s works, they will not be able to know why the objects are deleted

So it is a bit of extra work to write the code to remove the data, but it will be very explicit and it will be easy to debug/verify if something is not working as expected.

ziccardi:
👀

pkg/mobile/bindableServices.go
```diff
@@ -48,15 +89,32 @@ func (lister *BindableMobileServiceCRUDLImpl) Delete(namespace string, bindingNa
...
+	}
+
+	bindings, err := bms.scClient.ServiceBindings(mobileClient.GetNamespace()).List(bindListOpts)
+
+	if err != nil {
+		return err
+	}
+
+	for _, binding := range bindings.Items {
+		bms.Delete(mobileClient.GetNamespace(), binding.GetName())
```
jhellar: Not sure about this, but should we check for errors here? In this code [1] we return if there is an error - I thought that that is so that we won't delete an app if some of it's resources can not be deleted.

[1] https://github.com/aerogear/mobile-developer-console/pull/142/files#diff-891bcbc429e4f6564bef6d4f9d6abea4R21

wei-lee: ok, I will fix this.

PR #144, opened 2018-11-28
Fixed watching for resources in UI

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8149

## What

- watch for bindings in `Client` component (higher in hierarchy) rather than in `MobileServiceView`
- watch also for services in `Client`
- only show spinner in `MobileServiceView` for first fetch (not for every update)
- don't watch for services in Overview page

## Verification Steps - `watch for services`

1. Create app in MDC and open it
2. Provision a mobile service in OpenShift
3. Open Services tab in MDC
4. When the service is provisioned it should show up

## Verification Steps - `errors in console when app is deleted`

1. Create app in MDC and bind it with a service
2. Delete the app
3. There should be no error in browser console

### DIFF
--- a/ui/src/components/mobileservices/BoundServiceRow.js
+++ b/ui/src/components/mobileservices/BoundServiceRow.js
@@ -52,12 +52,11 @@ class BoundServiceRow extends Component {
       documentationFragment = (
         <Row>
           <Col md={2} className="detailsKey">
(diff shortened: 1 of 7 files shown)


psturc:
👀

PR #150, opened 2018-12-05
Fix: create mobile app Enter key forbidden

wojta:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8208

## What
Blocking Enter key in create app dialog.

## Why
When you pressed it, it would refresh the page without creating the app.

## How
Blocking keypress event in the app name field.

## Verification Steps
 
1. Go to overview screen
2. Press `Create Mobile App` 
3. Fill in name
4. Press Enter, it's expected for the dialog not to close.

## Checklist:

## Progress

- [x] Finished task


aliok:
:eye:

aliok:
@wojta the fix is valid.
but, can we simply create the app when I type something and press Enter? Like, trigger the Create button?

Anyway feel free to do that or not. I think we have this UX problems in so many other dialogs that pressing enter won't proceed the form.

wei-lee:
I am going to merge this one now and I will create a separate PR to improve this one.

PR #151, opened 2018-12-05
Update UI of app overview screen tiles

ziccardi:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8171

## What

- Added created label with creation date
- Moved services icons inside the card (centered vertically and horizontally)
- Moved build info to the footer

## Screenshot
screenshot

### DIFF
--- a/ui/src/components/overview/MobileClientCardViewItem.js
+++ b/ui/src/components/overview/MobileClientCardViewItem.js
@@ -42,7 +42,20 @@ const MobileClientCardViewItem = props => {
               <h1>{app.spec.name}</h1>
             </div>
 
-            <CardTitle>
+            <CardTitle />
+          </CardHeading>
+
+          <CardBody>
+            <div className="card-icons">
+              {services && services.length > 0 ? getServiceIcons(services) : <div className="service-icon" />}
+            </div>
+          </CardBody>
+          <CardFooter>
+            <div className="creation-timestamp">CREATED</div>
+            <span className="creation-timestamp">
+              <span className="fa fa-globe" /> <Moment format="DD/MM/YYYY">{app.metadata.creationTimestamp}</Moment>
+            </span>
+            <span className="builds">
               {buildTabEnabled && builds.numFailedBuilds > 0 ? (
                 <span>
                   <span className="pficon pficon-error-circle-o" />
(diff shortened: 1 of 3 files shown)


wei-lee:
when I create a new app, by default it looks like this:

screen shot 2018-12-06 at 11 08 21

I believe the extra boarder should not be there.

ui/src/components/overview/MobileClientCardViewItem.js
```diff
@@ -30,42 +31,47 @@ const MobileClientCardViewItem = props => {
...
+              {services && services.length > 0 ? getServiceIcons(services) : <div className="service-icon" />}
             </div>
           </CardBody>
           <CardFooter>
-            <div className="card-footer-icons">
-              {services && services.length > 0 ? getServiceIcons(services) : <div className="service-icon" />}
-            </div>
+            <div className="creation-timestamp">CREATED</div>
+            <span className="creation-timestamp">
+              <span className="fa fa-globe" /> <Moment format="DD/MM/YYYY">{app.metadata.creationTimestamp}</Moment>
```
wojta: Shouldn't we match locale of browser/serverside for date formatting? Or are we using US weird date format :)

ui/src/components/overview/MobileClientCardViewItem.js
```diff
@@ -30,42 +31,47 @@ const MobileClientCardViewItem = props => {
...
+                </span>
+              ) : null}
+              {buildTabEnabled && builds.numInProgressBuilds > 0 ? (
+                <span>
+                  <span className="pficon fa fa-refresh fa-spin fa-fw" />
+                  {builds.numInProgressBuilds}
+                </span>
+              ) : null}
+              {!buildTabEnabled || (builds.numFailedBuilds === 0 && builds.numInProgressBuilds === 0) ? <span /> : null}
+            </CardTitle>
```
wei-lee: According to the flats, the builds info should be moved to the footer. It's ok if we don't implement it for now, let's comment this one out, and add a todo here.

ui/src/components/overview/MobileClientCardViewItem.js
ziccardi: Not sure about that: none of our messages is localized, so I don't think we should localize the dates. WDYT?

wei-lee: yeah, let's not worry about localisation yet.  If we are going to do it, it will be part of the work to  support i18n in the UI.

wojta: ok for me

ui/src/components/overview/MobileClientCardViewItem.js
ziccardi: Moved the build info to the footer

ziccardi:
@wei-lee mind giving another look?

PR #153, opened 2018-12-05
 🐛 fix errors when try to create new build config

wei-lee:
## Motivation

Fix the errors when try to create new build configs.

## Verifications:

1. Enable build tab if it's not showing (you can change it via env vars)
2. Try create new build configs via the button or from the dropdown. They all should work.


aliok:
Looks good @wei-lee 
When I was getting rid of the appId in the MDC, @wojta  and I decided that we pass anything to build tab related stuff because those parts are out of scope for the release.
But glad to have them fixed.

PR #155, opened 2018-12-06
 🔖 use release tags in asb config

wei-lee:
## Motivation

Use the released tags in the ASB config


scripts/yml/aerogearcatalog_registry.yaml
```diff
@@ -3,8 +3,15 @@ registry:
     name: ag
     url: https://registry.hub.docker.com
     org: aerogearcatalog
-    tag: latest
+    tag: 1.0.0
     white_list:
       - '.*-apb$'
     black_list:
-      - '.*-app-apb$'
```
wei-lee: the app apbs are removed from docker hub, so we don't need the blacklist filter anymore

PR #158, opened 2018-12-07
Added monitoring of services to automatically update bindable services

ziccardi:
## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-8232

## What
The GUI was not updated automatically to reflect the list of available services (needed to manual refresh) when creating/deleting services from openshift console

## Changes
Added websocket watcher to check if the list of services was changed

## Step to verify
1. create a mobile app
1. open it
1. navigate to services tab
1. in another browser tab open OpenShift console and provision a new service
1. go back to mdc tab
1. check that the new service is visible into the MDC page
1. Try the same by deleting a service

### DIFF
--- a/ui/src/components/mobileservices/MobileServiceView.js
+++ b/ui/src/components/mobileservices/MobileServiceView.js
@@ -5,6 +5,8 @@ import { partition } from 'lodash-es';
 import BoundServiceRow from './BoundServiceRow';
 import UnboundServiceRow from './UnboundServiceRow';
 import './MobileServiceView.css';
+import DataService from '../../DataService';
+import { fetchBindings } from '../../actions/serviceBinding';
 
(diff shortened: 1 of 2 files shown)


wei-lee:
👀

wei-lee:
verified locally. Working as expected.

wei-lee:
while this does work, but I am curious that we did watch for services in here: https://github.com/aerogear/mobile-developer-console/blob/master/ui/src/containers/Client.js#L73. It is the parent contains of `MobileServicesView`, but it doesn't see to be working as expected.

ziccardi:
@wei-lee The problem is that the config panel is not using the `services` list to build the list of available services. In the parent, when services changes it just calls `fetchServices`. In the config panel we need the `fetchBindings` function to be called.

PR #161, opened 2018-12-11
Adding PKCS#12 validation

ziccardi:
## Motivation
Jira: https://issues.jboss.org/browse/AEROGEAR-8238

## What
- Removed the password confirmation field since it is not needed anymore (the password is verified agains the PKCS#12 file)
- Validators have been changed so that they won't depend anymore on the caller. 
- Two new validators has been added:
    - SameValueOf: used to check that a field has a specific value. That value can be the value of another field or the result of a passed in function. Used in this case for the password confirmation field.
    - P12Validator: used to check that tha passed in Base64 data is a PKCS#12 file and that the provided password is correct
- Validation config (`ValidationRules.json`) has been updated to enable the validation of the passed in PKCS#12 
- Added new unit tests


ziccardi:
@secondsun @StephenCoady Now that the PKCS#12 is validated against the provided password, I'm not sure it still has some use to have the password confirmation. WDYT?

ui/src/components/mobileservices/bindingPanelUtils.js
```diff
@@ -75,6 +75,39 @@ function getOpenShiftField(field, properties, idSchema, schema, formData, uiSche
...
+      <input
+        className="form-control"
+        type="password"
+        id={id}
+        onBlur={event => property.content.props.onChange(event.target.value)}
+      />
+    </div>
+  );
+
+  if (confirmation) {
```
jhellar: This might not be needed anymore.

ziccardi: I left it since the Template object can be used for other pages as well. In other pages the password confirmation field could make sense.

ziccardi:
@jhellar The issue should be resolved. Mind giving another look?

psturc:
👀

psturc:
@ziccardi have you seen this error before?

screen shot 2018-12-18 at 12 11 51

We hit this one after verifying the changes for UPS. This happened when I tried to bind metrics to a client.

ziccardi:
@psturc : giving a look 👀

PR #164, opened 2018-12-13
Fix for Input text fields gets emptied when other binding is in progress

jhellar:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8248

## What
`BindingPanel` moved from service rows to `ServiceView`.

## Why
Seems that BindingPanel gets rerendered, when there is an update from websocket.


psturc:
👀

PR #166, opened 2018-12-13
AEROGEAR-8239

ziccardi:
## Motivation
JIRA:

- https://issues.jboss.org/browse/AEROGEAR-8239
- https://issues.jboss.org/browse/AEROGEAR-8247

# What

- Fixed the wizard so that wizard state is always reset to initial state when opening
- Fixed platform filtering in filling the selectbox
- Fixed method used to list the bound/unbound services to comply with the rest of the MDC

# Verification steps

1. Create a new app
2. Bind UPS for iOS to the app
3. Try binding another UPS instance: only android must be available
4. After binding android too, try binding another service: the listbox should be empty (the bind button should disappear. Issue : [AEROGEAR-8273](https://issues.jboss.org/browse/AEROGEAR-8273)
5. Delete the iOS or the Android binding: only the binding for the other platform should be available
6. Delete the last UPS binding: binding for both platforms should be available again

### DIFF
--- a/ui/src/components/mobileservices/BindingPanel.js
+++ b/ui/src/components/mobileservices/BindingPanel.js
@@ -1,3 +1,4 @@
+import { partition, find } from 'lodash-es';
 import React, { Component } from 'react';
 import { Wizard } from 'patternfly-react';
 import { connect } from 'react-redux';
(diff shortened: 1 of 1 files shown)


ui/src/components/mobileservices/BindingPanel.js
```diff
@@ -77,37 +78,27 @@ export class BindingPanel extends Component {
...
-
-    for (const configExtItem of configExtItems) {
-      for (const variantInfo of configExtItem) {
-        if (variantInfo.type === platform) {
-          return true;
-        }
-      }
-    }
-    return false;
+  hasUPSIOSBoundServiceForPlatform(platform) {
```
jhellar: should this be `hasUPSBoundServiceForPlatform` instead of `hasUPSIOSBoundServiceForPlatform`?

ui/src/components/mobileservices/BindingPanel.js
```diff
@@ -57,6 +36,28 @@ export class BindingPanel extends Component {
...
+      schema,
+      form,
+      loading: false,
+      service,
+      activeStepIndex: 0
+    });
+  }
+
+  componentWillUpdate(nextProps, nextState) {
+    if (nextProps.showModal === false) {
```
jhellar: This has changed with my PR - `showModal` is never `false`.

ui/src/components/mobileservices/BindingPanel.js
```diff
@@ -205,7 +196,42 @@ export class BindingPanel extends Component {
...
+        }
+      } else if (hasAndroid && hasIOS) {
+        // UPS, there's already an IOS variant
+        if (schema.properties.CLIENT_TYPE) {
+          schema.properties.CLIENT_TYPE.default = '';
+          schema.properties.CLIENT_TYPE.enum = [];
+        }
+      }
+      // we don't care if there are variants for both platforms.
+      // this binding panel shouldn't be shown anyway
```
jhellar: do we need the `else if (hasAndroid && hasIOS)` case then?

ui/src/components/mobileservices/BindingPanel.js
ziccardi: 👍

ui/src/components/mobileservices/BindingPanel.js
ziccardi: It won't be needed anymore when AEROGEAR-8273 will be resolved

PR #168, opened 2018-12-17
Fixed windows automatically closing when binding finishes

ziccardi:
## Motivation
JIRA: https://issues.jboss.org/browse/AEROGEAR-8280

When creating a new binding, the create binding window did not close automatically when the binding was done


psturc:
👀

PR #175, opened 2019-01-09
feat: install yq if docker not present

psturc:
Verification:
1. Comment out line with [post install script](https://github.com/aerogear/mobile-developer-console/blob/master/scripts/minishift.sh#L24)
2. Run `./scripts/minishift.sh`
3. Stop docker app
4. Run `./scripts/post_install.sh`

### DIFF
--- a/scripts/install_yq.sh
+++ b/scripts/install_yq.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+cd "$(dirname "$0")" || exit 1
+
+source=""
+
+if [[ "$OSTYPE" == "linux-gnu" ]]
+then
+    source="https://github.com/mikefarah/yq/releases/download/2.2.1/yq_linux_amd64"
+elif [[ "$OSTYPE" == "darwin"* ]]
+then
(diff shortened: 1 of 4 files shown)


scripts/install_yq.sh
```diff
@@ -0,0 +1,22 @@
...
+else
+    echo "Error: could not detect os system type to download yq binary"
+    echo "Download suitable yq binary from https://github.com/mikefarah/yq/releases/tag/2.2.1 and add it to your path"
+    exit 1
+fi
+
+wget $source -O yq
+chmod +x ./yq
+sudo mv ./yq /usr/local/bin/
+echo "yq binary downloaded and moved to /usr/local/bin/"
```
wei-lee: on RHEL, the PATH is set to: `/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin:/root/bi`, there is no `/usr/local/bin/` on RHEL.

psturc: fixed by using `yq` directly from MDC folder.

PR #177, opened 2019-01-30
feature: Display mdcName label on service instance on bindings page 

secondsun:
## Motivation
There was a request to be able to customize the detail name of provisioned service instances. This PR enables that

## How
provisioned services may label their service instance with a label "mdcName" this label will be used on the binding rows as more detail.

## Verification Steps
1. Deploy a sync app and set the sync app name
2. Confirm that the detail on the service binding page is the sync app name and not "Sync App"


psturc:
Specifically I meant some links in "Bound services" section. See the screenshot: Sync App compared to Metrics and Keycloak.

screenshot 2019-02-04 at 12 11 07

secondsun:
@psturc I'm trying to make the certificate situation with the apb's less awful, I'll touch this one as well.
Is this pr ok to merge?

psturc:
Yep, already approved. Those changes need to be done in apb anyway [see e.g. Keycloak](https://github.com/aerogearcatalog/keycloak-apb/blob/master/roles/bind-keycloak-apb/tasks/main.yml#L124-L131)

wei-lee:
@secondsun what's the story with this PR, is it required, should we merge it?

secondsun:
@wei-lee Chris asked for it during the demo before last.  It should have been merged a long ass time ago.  It gives sync apps pretty names in the mdc console

PR #190, opened 2019-02-27
Issue AEROGEAR-8604

ziccardi:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-8604

## What
- Added a check to avoid crashing in case of unknown service types
- Added unit tests to verify this specific issue

## Why
When bound, the sync service app was expected to be of type `sync` instead of `sync-app` (see [here](https://github.com/aerogearcatalog/sync-app-apb/blob/master/roles/bind-sync-app-apb/tasks/main.yml#L42)).

## Verification Steps
 
1. Create a new app in MDC
2. Provision the sync-app into the MDC project
3. Bind the sync-app to the app you created in step 1
4. Open the app details, go to the configuration tab and click on ionic and/or cordova

Without this fix, it will crash. With this fix, an error will be logged into the console.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,8 @@ jobs:
       - run: go get -u github.com/golang/dep/cmd/dep
       - run: make build
       - run: make test_cover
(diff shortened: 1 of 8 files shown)


psturc:
👀

PR #231, opened 2019-04-17
[WIP] feat: add a nodejs server and refactor app CRUDL endpoints

wei-lee:
## Motivation

This will be a long running branch to contain the refactoring work of MDC.

This PR:

* Adds a new nodejs server
* Refactored the apps CRUDL & watch endpoints
* Add script for local development

To verify:

* Start a new minishift server and make sure the mobileclients CRDs & roles are setup correctly
* Run `local_dev.sh`
* In the console, try create, edit, delete apps.


wei-lee:
@jhellar @ziccardi feel free to take a look at this PR and give your feedbacks.

ui/local_dev.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+MINISHIFT_IP=`minishift ip`
+USER_TOKEN=`oc whoami -t`
+
+if [ -z "$MINISHIFT_IP"]; then
```
ziccardi: This does not work on my machine. When minishift is not running, the content of `MINISHIFT_IP` is `Running this command requires a running 'minishift' VM, but no VM is running.`

I changed it to `minishift status | head -1 | awk -F: '/ /{gsub(/ /,  "", $2); print $2}'`: this returns `Stopped` if minishift is stopped, `Running` otherwise

ui/server.js
```diff
@@ -0,0 +1,142 @@
...
+      accessToken: '${userToken}',
+      name: '${userName}',
+      email: '${userEmail}'
+    }
+  };`;
+}
+
+function getServices(servicesConfigPath) {
+  return new Promise(resolve => {
+    if (fs.existsSync(servicesConfigPath)) {
```
ziccardi: Not sure this has really a big value. We could remove this check and lines 103-104 and get almost the same result.

jhellar:
Verified operations on mobile clients - looks good.

Dockerfile
```diff
@@ -1,9 +1,13 @@
-FROM centos:7.4.1708
-ARG BINARY=./mobile-developer-console
+FROM registry.access.redhat.com/rhoar-nodejs/nodejs-10
```
grdryn: registry.access.redhat.com is due to go away at some point (either this year or next year I'd say) so maybe we shouldn't use it? Also, if we intend to build this image and host it somewhere like quay.io, I'm not sure if we'd be allowed (it depends on the base of this image I think). I'm not sure what alternative to use instead, do they have something on quay.io?

wei-lee: I don't really know TBH, couldn't find any other documents. Maybe we should just use the community nodejs image?

grdryn: That would be a better option I think. Alternatively there's this image which is the upstream version of the RHSCL Node.js 10 image I believe: https://hub.docker.com/r/centos/nodejs-10-centos7

I wasn't able to find any upstream image from RHOAR, but I might just not be looking in the right places. I'll see if I can find someone to ask

PR #237, opened 2019-05-13
delete golang server

wei-lee:


.circleci/config.yml
```diff
@@ -3,93 +3,58 @@
...
+      - run: make test-cover
+      - run: make build
 
 
   docker_push_master:
-    working_directory: /go/src/github.com/aerogear/mobile-developer-console
+    working_directory: ~/mobile-developer-console
     docker:
-      - image: circleci/golang:1.10
+      - image: circleci/node:8.15.0
```
grdryn: Should/could we go straight to Node.js 10?

wei-lee: probably should. Will try it.

PR #250, opened 2019-05-29
Aerogear 9291 data sync

wei-lee:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9291

## Verification Steps

1. Run this branch. Create a new mobile app or use an existing one.
2. Go to the services page and you should see there is a `Data Sync` service.
3. Bind the `Data Sync` service. Fill the values in the binding step.
4. Once it's finished, go to the configuration page and you should see the content of `mobile-services.json` file is updated.
5. Go back to the services page and delete the binding for data sync
6. Check the content of `mobile-services.json` file and it should be updated again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ npm-ci:
 
 .PHONY: check-code-style
 check-code-style: setup
-	npm run build-css && node node_modules/eslint/bin/eslint src/
+	npm run build-css && npm run lint
 
 .PHONY: test-cover
 test-cover: npm-ci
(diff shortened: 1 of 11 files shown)


craicoverflow:
I'm still seeing this error message that was also present in #241. I assumed at the time this was because not all of the services were created yet, but now that there are 2 and they are working it must be coming from somewhere else.

I tried sourcing the error and I believe it comes from [`src/actions/fetch.js`](https://github.com/aerogear/mobile-developer-console/blob/c151bb4532b63358224f41bee47a4ef78aa55850/src/actions/fetch.js) as it is getting added to the `ToastNotificationList` in the `ErrorMessages` component.

Screenshot from 2019-05-30 09-36-02

wei-lee:
@craicoverflow Create a JIRA for it and we will fix it in a separate PR.

wei-lee:
@craicoverflow I have addressed your comments now.

PR #253, opened 2019-05-30
setup keycloak service during prepare step

wei-lee:
## Motivation

Update the `prepare.sh` script to provision the services.
 
To verify:
* run `prepare.sh` script and it should provision keycloak into a namespace

### DIFF
--- a/scripts/prepare.sh
+++ b/scripts/prepare.sh
@@ -3,11 +3,31 @@
 cd "$(dirname "$0")" || exit 1
 
 # You only need to run this script once to setup a cluster
+NAMESPACE="${NAMESPACE:-mobile-console}"
 MOBILE_CLIENT_CRD="${MOBILE_CLIENT_CRD:-../deploy/mobile-client-crd.yaml}"
 OAUTH_CLIENT_FILE="${OAUTH_CLIENT_FILE:-../deploy/openshift-oauthclient.yaml}"
 OC_CMD="${OC_CMD:-oc}"
 
+echo "Create project ${NAMESPACE}"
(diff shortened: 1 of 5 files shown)


scripts/prepare.sh
```diff
@@ -3,11 +3,31 @@
 cd "$(dirname "$0")" || exit 1
 
 # You only need to run this script once to setup a cluster
+NAMESPACE="${NAMESPACE:-mobile-console}"
```
craicoverflow: If a developer follows [these steps](https://github.com/aerogear/mobile-developer-console#local-openshift-cluster) in the README:

```sh
./scripts/minishift_start.sh
./scripts/prepare.sh
./scripts/development.sh
```
`prepare.sh` will set everything up in the `mobile-console` namespace, but `development.sh` switches to `myproject`. I think this script should also be updated with the same namespace or `oc project myproject` could be removed altogether because it will already be in the correct namespace.

wei-lee: yes, good point. I will update it.

wei-lee:
@craicoverflow mind checking again?

craicoverflow:
> @craicoverflow mind checking again?

For me, `development.sh` switches to the `mobile-console` namespace, but when the development server runs it still uses "myproject" for API calls.

Once I added `export NAMESPACE="${NAMESPACE}"` again to `development.sh` "mobile-console" is used by the server.

wei-lee:
@craicoverflow ok, I have exported the namespace as part of the script now. Also the server is updated to  print out a warning if namespace is not set and which default namespace will be used.

PR #254, opened 2019-05-30
make sure set the owner references field on the resources

wei-lee:
## Motivation

Fix the following issues:

1. Error in [this comment](https://github.com/aerogear/mobile-developer-console/pull/250#issuecomment-497253878)
2. If an app is bound to services, when the app is deleted and re-created, the bound services are still there.

To verify:

1. Click on an app from the overview page, and there shouldn't be error reported anymore.
2. Bind a service to the app, delete the app, recreate it again with the same name and you should see there are no bound services for the app.

### DIFF
--- a/src/components/mobileservices/BindingPanel.js
+++ b/src/components/mobileservices/BindingPanel.js
@@ -125,7 +125,7 @@ export class BindingPanel extends Component {
   stepChanged = step => {
     if (step === 2) {
       this.setState({ loading: true });
-      this.props.createCustomResourceForService(this.state.service, this.state.formData);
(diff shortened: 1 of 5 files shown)


src/components/mobileservices/BindingPanel.js
```diff
@@ -184,8 +184,10 @@ const mapDispatchToProps = {
   createCustomResourceForService
 };
 
-function mapStateToProps(state) {
+function mapStateToProps(state, ownProps) {
```
craicoverflow:
```suggestion
function mapStateToProps(state, { appName }) {
```

PR #256, opened 2019-05-31
update the services list for apps in the background

wei-lee:
## Motivation

Previously the content for `mobile-services.json` file is generated when a request is sent to an endpoint on the server. However, there are a few problems with this approach:

1. Turns out it doesn't work when using a cli. The MDC is protected by the oauth-proxy and it doesn't allow requests to pass through without a cookie. But a cookie can only be generated by perform a login against the oauth-proxy, which is tricky to do.
2. Introduces a few issues in the UI: the icons for bound services are not showing up on the overview page, the docs on the configuration page is not updated correctly etc.

To solve these problems, the best way is implementing what it was done before in the old golang server: update the `status` field of the mobileclients with the content in the background. This way user will be able to:

1. Use `oc` to get the content of mobile-services.json file by inspecting the output of a mobileclient: `oc get mobileclient app1 -o json`
2. The bugs in the UI are fixed automatically without other changes.

to verify:

Try bind & unbind for the app and everything should still work as before. When there are services bound for an app, you should also see the icons shows up for the app on the overview page. If you look in to the docs section on the configuration page, you should also see that the content is updated depending on the bound services for the app.


wei-lee:
@craicoverflow thanks for the review. I have addressed them now. Mind taking another look?

craicoverflow:
I did not notice this last time - IDM is missing its icon on the Configuration section. It is trying to load a PNG, but the Keycloak icon is an SVG. Once this is fixed I will approve.

Screenshot from 2019-05-31 13-38-22

wei-lee:
@craicoverflow fixed. If you are happy with PR, once approved you can merge it.

PR #261, opened 2019-06-04
test: added tests for BuildInformation component

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9325

## What

Added a test file for the `BuildInformation` component.

## Why

Previously there were none.

## Verification Steps
 
1. Run `npm run coverage | grep 'Lines\|BuildInformation.js'`
2. The coverage level should be at 100%.

### DIFF
--- a/src/components/build/BuildInformation.test.js
+++ b/src/components/build/BuildInformation.test.js
@@ -1,0 +1,65 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import BuildInformation from './BuildInformation';
+
+const setup = (propOverrides = {}) => {
+  const defaultProps = {
+    build: {},
+    appName: ''
+  };
+
+  const props = { ...defaultProps, propOverrides };
+  const wrapper = shallow(<BuildInformation {...props} />);
+
+  return { props, wrapper };
+};
+
+describe('BuildInformation', () => {
+  const {
+    props: { build },
+    wrapper
+  } = setup({
+    appName: 'myapp',
+    build: {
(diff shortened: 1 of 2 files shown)


src/components/build/BuildInformation.test.js
```diff
@@ -0,0 +1,65 @@
...
+      expect(BuildPipelineStage).toHaveLength(1);
+      expect(BuildPipelineStage.props().build).toEqual(build);
+    });
+  });
+
+  describe('BuildDownload', () => {
+    it('should match previous snapshot', () => {
+      const BuildDownload = wrapper.find('BuildDownload');
+
+      expect(BuildDownload).toMatchSnapshot();
```
craicoverflow: Since `BuildDownload` is a container component which is connected to the store, I was not sure of the bet way to test it from here, since it does not act like regular components. 

This compares the JSON output of the component to the previous time and fails if they do not match.

If you think this is a bad approach please let me know and I will look at changing it.

wei-lee: The pattern we have used in MDC to test container component is:

1. Export both components (the one that is connected to the store, and the one that isn't) from the container class.
2. In the test, only test the component that is not connected to the store.

A good example is this: https://github.com/aerogear/mobile-developer-console/blob/master/src/containers/Client.test.js.

This way you can still test the component without building a store.

craicoverflow: Thanks.

PR #264, opened 2019-06-06
fix: missing oauth client

psturc:
## Motivation
By following the steps here, a user gets this error if a new cluster is used to deploy MDC for the first time:
```
Error from server (NotFound): oauthclients.oauth.openshift.io "mobile-developer-console" not found
```

## Verification Steps
1. Target a new remote OpenShift cluster
2. Follow the steps from [here](https://github.com/aerogear/mobile-developer-console#remote-cluster)
3. Make sure that mdc is deployed successfully and you can log in to MDC as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,15 +22,21 @@
     ```
     oc login <openshift-url>
     ```
-2. Deploy the service using one of the following commands:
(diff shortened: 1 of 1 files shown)


wei-lee:
@psturc I think the real problem here is that we should update the README to ask users to run the `prepare.sh` first. There are other prerequisites are missing I think. Remember it is it not required to run the installer anymore.

psturc:
@wei-lee I've updated the readme file

PR #284, opened 2019-06-21
feat: use metadata.name instead of spec.name

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9446

## What

1. Remove the ability to edit an app.
2. Use `metadata.name` instead of `spec.name` for the MDC UI.
3. Fixed a bug where the `status` object will not full create sometimes.

## Why

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/aerogear/UelPDpo2tww

## Verification Steps
 
1. Create an app in MDC.
2. Try to edit the app. You should not be able to.
3. Edit `spec.name` of the app resource in OpenShift.
4. The name change should not affect the display name in MDC.
5. Verify that the status object is populated with `clientId`, `services` (empty array) and `namespace`.

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/deploy/mobile-client-crd.yaml
+++ b/deploy/mobile-client-crd.yaml
@@ -21,6 +21,7 @@ spec:
             apiKey:
               type: string
               pattern: '(\w{8}-\w{4}-\w{4}-\w{4}-\w{11})'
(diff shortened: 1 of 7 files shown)


ziccardi:
👀

PR #286, opened 2019-06-25
Aerogear 9471

craicoverflow:
## Motivation
Add references to relevant tickets or a short description about what motivated you do it. (E.g JIRA: https://issues.jboss.org/browse/AEROGEAR-{} AND/OR ISSUE: https://github.com/aerogear/standards/issues/{}) 

## What
Add an short answer for: What was done in this PR? (E.g Don't allow users has access to the feature X.)

## Why
Add an short answer for: Why it was done? (E.g The feature X was deprecated.)

## How
Add an short answer for: How it was done? (E.g By removing this feature from ... OR By removing just the button but not its implementation ... ) 

## Verification Steps
 
1. Go to `XX >> YY >> SS`
2. Create a new item `N` with the info `X`
3. Try to edit this item 
4. Check if in the left menu the feature X is not so long present.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes


server/appServices.js
```diff
@@ -85,6 +91,52 @@ function updateAppsAndWatch(namespace, kubeclient) {
...
+  const jsonStream = new JSONStream();
+  stream.pipe(jsonStream);
+
+  jsonStream.on('data', event => {
+    if (event.type === 'ADDED') {
+      DeviceSecurityService.disabled = false;
+    } else if (event.type === 'DELETED') {
+      DeviceSecurityService.disabled = true;
+    } else if (event.type !== 'MODIFIED') {
+      watchMobileSecurityServiceCR(namespace, kubeclient);
```
craicoverflow: @ziccardi - how familar are you with the `kubeclient`? For some reason, after a few times of installing and uninstalling the Mobile Security Service CR and namespace, then `event` becomes a string "/version", before hand it was an object.

This is the only way I have found to get it working, but I don't think it is a good approach as it is adding extra watches to the stream. Any ideas?

ziccardi:
👍

PR #290, opened 2019-06-28
Implementing custom resource manager

ziccardi:
## Motivation
The creation of a PushVariantCR is different than what we had until now: before we can create a variant, we have to check that an application already exists. It it doesn't it must be created. Then the new variant must be bound to this application.

## What
Added a resource manager factory so that a different resource manager can be returned for different resources. It defaults to the GenericResourceManager whose behaviour is the one we had until now.

## Verification Steps
 
## Checklist:

### DIFF
--- a/src/services/crmanagers/OpenshiftResourceManagerFactory.js
+++ b/src/services/crmanagers/OpenshiftResourceManagerFactory.js
@@ -1,0 +1,27 @@
+import { GenericResourceManager } from './GenericResourceManager';
+import { PushVariantResourceManager } from './PushVariantResourceManager';
+import { getUser } from '../openshift';
+
+class UserBoundResourceManager {
+  constructor(resourceManager) {
+    this._resourceManager = resourceManager;
+  }
+
+  create = (res, obj, owner) => getUser().then(user => this._resourceManager.create(user, res, obj, owner));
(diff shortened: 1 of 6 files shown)


src/services/crmanagers/GenericResurceManager.js
```diff
@@ -0,0 +1,159 @@
...
+
+  catch(handler) {
+    this._errorHandler = handler;
+    return this;
+  }
+}
+
+/* eslint-disable class-methods-use-this */
+
+export class GenericResurceManager {
```
craicoverflow: Typo: Should be `GenericResourceManager`.

src/services/crmanagers/OpenshiftResourceManagerFactory.js
```diff
@@ -0,0 +1,12 @@
+import { GenericResurceManager } from './GenericResurceManager';
+import { PushVariantResourceManager } from './PushVariantResourceManager';
+
+export class OpenshiftResourceManagerFactory {
```
craicoverflow: I believe this could simply be called `ResourceManagerFactory`. What do you think?

ziccardi:
Embedded into #306

PR #300, opened 2019-07-03
Aerogear 9539

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9539

## What
Added validation around "Create build config" name field.

## Why
To reduce the risk of a user getting a 422 response back from the server

## How
This was achieved by adding Regex checking of the "Create build config" name. This uses the same regex expression that checks the apps name on creation. 

## Verification Steps
1. Enable build tab by setting the ENABLE_BUILD_TAB environment variable to true when running MDC.
2. Open the Build tab and click Create build config.
3. Enter a value that should be invalid in the name field. For example - "hhh!!!!www.".
4. Fill out the rest of the form as normal.
5. You should see a red highlighting around the name field and the Create button should be inactive.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

image

### DIFF
--- a/src/components/build/create_build_config/Validations.js
+++ b/src/components/build/create_build_config/Validations.js
@@ -18,7 +18,13 @@ import {
   KEY_CR_BUILD_IOS_CREDENTIALS_PROFILE_PASSWORD,
   withPath
 } from '../Constants';
-import { VALIDATION_OK, validateNotEmpty, validateGitUrl, VALIDATION_ERROR } from '../../common/Validation';
+import {
+  VALIDATION_OK,
+  validateNotEmpty,
+  validateGitUrl,
+  VALIDATION_ERROR,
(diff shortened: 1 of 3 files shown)


src/components/build/create_build_config/Validations.js
```diff
@@ -60,9 +66,12 @@ export function checkMandatoryFields(state) {
 }
 
 export function configValidation(key, value) {
+  let response = {};
   switch (key) {
     case KEY_CR_NAME:
-      return validateNotEmpty(value);
+      response = validateNotEmpty(value);
```
craicoverflow: Since this is being overwritten immediately anyway it can be removed. `validateNameString` also fails if the name is empty so it catches both.

PR #304, opened 2019-07-04
test(actions/users.js): Tests added

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9547

## What
Adding tests to actions/users.js

## Why
To improve the number of tests

## How
Add unit tests to **actions/users.test.js**

## Verification Steps
1. Run `npm run test`
2. All tests should be passing

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/src/actions/users.test.js
+++ b/src/actions/users.test.js
@@ -6,31 +6,7 @@ import { ERROR } from './errors';
 const middlewares = [thunk];
 const mockStore = configureMockStore(middlewares);
 
-const userInfoRequest = 'USER_INFO_REQUEST';
-const userInfoSuccess = 'USER_INFO_SUCCESS';
-const userInfoFailure = 'USER_INFO_FAILURE';
-
-const result = {
-  accessToken: 'vlseTWA8LXFz6X5kBH2BJO2l33GRZSPZAHA2-FHGC94',
-  name: 'mockuser',
-  email: 'mockuser@example.com'
-};
-
-describe('All constants remain unchanged', () => {
-  it('USER_INFO_REQUEST == USER_INFO_REQUEST', () => {
-    expect(USER_INFO_REQUEST).toEqual(userInfoRequest);
-  });
-
-  it('USER_INFO_SUCCESS == USER_INFO_SUCCESS', () => {
-    expect(USER_INFO_SUCCESS).toEqual(userInfoSuccess);
-  });
-
-  it('USER_INFO_FAILURE == USER_INFO_FAILURE', () => {
-    expect(USER_INFO_FAILURE).toEqual(userInfoFailure);
-  });
-});
-
-describe('Test actions', () => {
+describe('Actions', () => {
   let store;
   let initialState;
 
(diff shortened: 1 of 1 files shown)


src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+  beforeAll(() => {
+    initialState = {};
+  });
+
+  beforeEach(() => {
+    store = mockStore(initialState);
+  });
+
+  describe('fetchUserInfo', () => {
+    it('successfully fetch user info', () => {
```
craicoverflow: As we are mainly testing that the correct actions are created here I think the description should focus on the actions we expect. So for this one it could be:

```suggestion
it('creates USER_INFO_SUCCESS action', () => {}
```

src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+      const expectedActions = [{ type: userInfoRequest }, { result, type: userInfoSuccess }];
```
craicoverflow: It is better to remove the variables `userInfoRequest` and `userInfoSuccess` that are defined at the top of this file and instead reference the action constants directly. It would also make this easier to read as looking at this line in isolation it is hard to know what the actual action is.

src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+
+      return store.dispatch(fetchUserInfo()).then(() => {
+        const actions = store.getActions();
+        expect(actions).toEqual(expectedActions);
+      });
+    });
+
+    it('window.OPENSHIFT_CONFIG.user = null', () => {
```
craicoverflow:
```suggestion
it('creates USER_INFO_FAILURE action', () => {}
```

src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+      const expectedActions = [
+        { type: userInfoRequest },
+        { error: new Error('no user found'), type: userInfoFailure },
+        { error: new Error('no user found'), type: ERROR }
+      ];
+
+      window.OPENSHIFT_CONFIG.user = null;
```
craicoverflow: This will cause problems in the future if another test is added below this one. `window.OPENSHIFT_CONFIG.user` has been set to null and has not been reset so a test that runs below this would not have access to the original user object.

You can use `beforeEach` to reset the user by making a clone of the `window.OPENSHIFT_CONFIG` in `beforeAll` (use ES6 spread operator for this).

https://jestjs.io/docs/en/setup-teardown.html

src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+import { ERROR } from './errors';
+
+const middlewares = [thunk];
+const mockStore = configureMockStore(middlewares);
+
+const userInfoRequest = 'USER_INFO_REQUEST';
+const userInfoSuccess = 'USER_INFO_SUCCESS';
+const userInfoFailure = 'USER_INFO_FAILURE';
+
+const result = {
```
craicoverflow: You can remove this as it is duplicating the user object defined in `window.OPENSHIFT_CONFIG`. This means it would need to be updated in two places.

src/actions/users.test.js
```diff
@@ -0,0 +1,70 @@
...
+  it('USER_INFO_SUCCESS == USER_INFO_SUCCESS', () => {
+    expect(USER_INFO_SUCCESS).toEqual(userInfoSuccess);
+  });
+
+  it('USER_INFO_FAILURE == USER_INFO_FAILURE', () => {
+    expect(USER_INFO_FAILURE).toEqual(userInfoFailure);
+  });
+});
+
+describe('Test actions', () => {
```
craicoverflow: You can remove "Test" as it is redundant.

PR #307, opened 2019-07-04
fix: return promise

craicoverflow:
Since the promises were not being returned the tests would pass when they should fail.


ziccardi:
Moved to #313

PR #315, opened 2019-07-10
test(SameValueOfRule): Basic tests added

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9561

## What
Add test coverage to SameValueOfRule.js

## Why
To increase the test coverage.

## How
Added SameValueOfRule.test.js to **components/mobileservices/validator/rules**

## Verification Steps
 
1. Run `npm run test`
2. All tests should be passing

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/src/components/mobileservices/validator/rules/SameValueOfRule.test.js
+++ b/src/components/mobileservices/validator/rules/SameValueOfRule.test.js
@@ -1,0 +1,69 @@
+import { SameValueOfRule } from './SameValueOfRule';
+
+describe('SameValueOfRule Validation', () => {
+  it('target value is a undefined', () => {
+    const config = { target: 'data.passwordConfirmation', error: 'Custom error message' };
+    const sameValueOfRule = new SameValueOfRule(config);
+
+    const input = {
+      data: { password: undefined, passwordConfirmation: undefined }
+    };
+    const value = sameValueOfRule.validate(input, 'data.password');
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
+
+  it('target value is a null', () => {
+    const config = { target: 'data.passwordConfirmation', error: 'Custom error message' };
+    const sameValueOfRule = new SameValueOfRule(config);
+
+    const input = {
+      data: { password: null, passwordConfirmation: null }
+    };
+    const value = sameValueOfRule.validate(input, 'data.password');
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
+
+  it('target value is a string', () => {
+    const config = { target: 'data.passwordConfirmation', error: 'Custom error message' };
+    const sameValueOfRule = new SameValueOfRule(config);
+
+    const input = {
+      data: { password: '12345678', passwordConfirmation: '12345678' }
+    };
+
+    const value = sameValueOfRule.validate(input, 'data.password');
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
+
+  it('target value is a function', () => {
+    const config = { target: () => '12345678', error: 'Custom error message' };
+    const sameValueOfRule = new SameValueOfRule(config);
+
+    const input = {
+      data: { password: '12345678', passwordConfirmation: '12345678' }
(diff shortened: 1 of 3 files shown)


src/components/mobileservices/validator/rules/SameValueOfRule.test.js
```diff
@@ -0,0 +1,72 @@
...
+    const sameValueOfRule = new SameValueOfRule(config);
+
+    const input = {
+      data: { app: undefined },
+      key: 'app'
+    };
+    const value = sameValueOfRule.validate(input.data, input.key);
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
```
ziccardi: The `SameValueOfRule` is used to check that a specific field has the same value of another field. This is usually used for fields like 'password & password confirmation' or 'mail & mail confirmation'.

This test gives a positive answer, but, IMO, it doesn't test the `SameValueOfRule` _purpose_ since in the actual code, the tests is checking that `input.data.app` equals `input.data[input.data.key]; however, since `input.data.key` is `app`, it is simply checking that `input.data.app` is equals to `input.data.app`

I believe it would be better to rewrite it as something like:
```javascript
  it('target value is a undefined', () => {
    const config = { target: 'data.passwordConfirmation', error: 'Custom error message' };
    const sameValueOfRule = new SameValueOfRule(config);

    const input = {
      data: { password: undefined, passwordConfirmation: undefined },

    };
    const value = sameValueOfRule.validate(input, 'data.password');

    expect(value).toEqual({ valid: true });
  });
```

src/components/mobileservices/validator/rules/SameValueOfRule.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+    const input = {
+      data: { app: null },
+      key: 'app'
+    };
+    const value = sameValueOfRule.validate(input.data, input.key);
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
```
ziccardi: See comment above

src/components/mobileservices/validator/rules/SameValueOfRule.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+    const input = {
+      data: { app: 'app' },
+      key: 'app'
+    };
+    const value = sameValueOfRule.validate(input.data, input.key);
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
```
ziccardi: For the same reason exposed above, I believe this should be something like :
```javascript
  it('target value is a string', () => {
    const config = { target: 'data.passwordConfirmation', error: 'Custom error message' };
    const sameValueOfRule = new SameValueOfRule(config);

    const input = {
      data: { password: '12345678', passwordConfirmation: '12345678' },
    };
    const value = sameValueOfRule.validate(input, 'data.password');

    expect(value).toEqual({ valid: true });
  });
```

src/components/mobileservices/validator/rules/SameValueOfRule.test.js
```diff
@@ -0,0 +1,72 @@
...
+
+    const input = {
+      data: { app: 'functionTest' },
+      key: 'app'
+    };
+    const value = sameValueOfRule.validate(input.data, input.key);
+    const expectResult = { valid: true };
+
+    expect(value).toEqual(expectResult);
+  });
```
ziccardi: See previous comments.

src/components/mobileservices/validator/rules/SameValueOfRule.test.js
```diff
@@ -0,0 +1,69 @@
+import { SameValueOfRule } from './SameValueOfRule';
+
+describe('SameValueOfRule Validation', () => {
+  it('target value is a undefined', () => {
```
craicoverflow: I think it would be better if the `it` text describes what should happen.

```suggestion
it('should be valid when target value is undefined', () => {
```

This goes for every `it` function in the test file.

PR #330, opened 2019-07-17
update the format of the service list

wei-lee:
## Motivation

Update the format of the service list file to match what is generated by the installer. See https://github.com/integr8ly/installation/pull/748/files#r304318416

## What

* Update the format of the services list and filter them using the `mobile:true` flag
* Remove the redundant `/metrics` endpoint. It is added by the Prometheus middleware automatically
* Fix the OPENSHIFT_HOST url if the protocol is specified as part of the env var 

## Verification Steps

* Run MDC and everything should still work as expected. 
* Add or remove the protocol value of the OPENSHIFT_HOST env var and it should still work.
* The `/metrics` endpoint should still work.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -116,10 +116,11 @@ function getConfigData(req) {
     userName = req.get('X-Forwarded-User');
     userEmail = req.get('X-Forwarded-Email');
   }
-  const parsedHost = new URL(process.env.OPENSHIFT_HOST);
-  if (!parsedHost.protocol) {
-    parsedHost.protocol = 'https';
+  let host = process.env.OPENSHIFT_HOST;
+  if (host && host.indexOf('http') === -1) {
+    host = `https://${host}`;
   }
+  const parsedHost = new URL(host);
   const masterUri = parsedHost.toString().slice(0, -1);
   parsedHost.protocol = 'wss';
   const wssMasterUri = parsedHost.toString().slice(0, -1);
--- a/server/server.js
+++ b/server/server.js
@@ -126,9 +126,9 @@ function getConfigData(req) {
   let host = process.env.OPENSHIFT_HOST;
   host = addProtocolIfMissing(host);
   const parsedHost = new URL(host);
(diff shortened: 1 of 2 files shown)


server/server.js
```diff
@@ -111,15 +113,23 @@ function getConfigData(req) {
     userName = req.get('X-Forwarded-User');
     userEmail = req.get('X-Forwarded-Email');
   }
+  let host = process.env.OPENSHIFT_HOST;
+  if (host && host.indexOf('http') === -1) {
```
craicoverflow:
```suggestion
if (host && host.startsWith('http') === -1) {
```

This may be faster, as it does not need to traverse the entire string. This is probably a bit of a micro-optimisation though.

server/server.js
```diff
@@ -32,38 +32,40 @@ app.use(
...
     //   type: MetricsService.type,
     //   url: `https://${process.env.METRICS_URL || process.env.OPENSHIFT_HOST}`
     // }
-    {
-      type: DataSyncService.type
-    },
     {
       type: MobileSecurityService.type,
-      url: `https://${process.env.MSS_URL || process.env.OPENSHIFT_HOST}`
+      host: `https://${process.env.MSS_URL || process.env.OPENSHIFT_HOST}`,
```
craicoverflow: If you set `OPENSHIFT_HOST` to include the protocol (e.g `https://example-url.com`) then this host will be set to `https://https://example-url.com` if `MSS_URL` is not set.

The same would occur with the other services that have a host key.

Screenshot from 2019-07-17 15-29-41 (copy)

server/server.js
craicoverflow:
```suggestion
if (host && host.startsWith('http') === -1) {
```

Probably a micro optimisation but `startsWith` would be faster as it does not need to traverse the entire string.

wei-lee:
@craicoverflow I have addressed your comments, mind taking another look?

PR #331, opened 2019-07-18
refactor(kubeclient): remove deprecation warnings

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9602

## What

Removed deprecation warnings for [godaddy/kubernetes-client](https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md) library.

## Verification Steps
 
1. Run the MDC server: `OPENSHIFT_HOST=$(minishift ip):8443 OPENSHIFT_USER_TOKEN=$(oc whoami -t) npm run start:server`
2. There will now only be 3 deprecation warnings printed in the console on startup:

```sh
kubernetes-client deprecated .getStream see https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md server/appServices.js:67:22                  
kubernetes-client deprecated .getStream see https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md server/appServices.js:97:24                  
kubernetes-client deprecated .getStream see https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md server/appServices.js:108:20
```

3. Run the MDC UI. MDC should still work as normal.

## Checklist:

## Additional Notes

Three deprecation warning still appear. These are related to streaming `watch` endpoints on custom resource. The documentation for using `getObjectStream()` (which removes this warning and works fine on default resources such as pods) does not work for streaming custom resources.

See [merging-with-kubernetes.md#getstream](https://github.com/godaddy/kubernetes-client/blob/master/merging-with-kubernetes.md#getstream).

I have an open issue https://github.com/godaddy/kubernetes-client/issues/517 about this.

### DIFF
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -66,32 +66,26 @@ function getServicesForApp(namespace, app, kubeclient) {
   return Promise.all(promises).then(serviceConfigs => serviceConfigs.filter(Boolean));
 }
 
-function updateAppsAndWatch(namespace, kubeclient) {
+async function updateAppsAndWatch(namespace, kubeclient) {
   updateAll(namespace, kubeclient).then(async () => {
     watchMobileClients(namespace, kubeclient);
 
-    const secretStream = kubeclient.api.v1.watch.namespace(namespace).secrets.getStream();
-    const secretsJsonStream = new JSONStream();
-    secretStream.pipe(secretsJsonStream);
-    secretsJsonStream.on('data', event => {
+    const secretStream = await kubeclient.api.v1.watch.namespace(namespace).secrets.getObjectStream();
+    secretStream.on('data', event => {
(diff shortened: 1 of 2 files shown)


craicoverflow:
@wei-lee we can maybe keep this open until next week in case a reply with a resolution is found with https://github.com/godaddy/kubernetes-client/issues/517.

wei-lee:
@craicoverflow can you still reproduce the issue that we have seen before?

craicoverflow:
> @craicoverflow can you still reproduce the issue that we have seen before?

Yes unfortunately the issue still occurs. I cannot find a way to watch custom resources using the new `getObjectStream()`. It would be nice to see if that would fix it.

craicoverflow:
@wei-lee I think we could close this one off and merge without removing the last 3 deprecation warnings as at least the majority of work is done and it doesn't affect how MDC runs.

PR #335, opened 2019-07-22
fix(kubeclient): keep watch open on CRDs

craicoverflow:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9607

## What

[godaddy/kubernetes-client](https://github.com/godaddy/kubernetes-client) watch closes after an indertminate amount of time, on **custom** resources only. This fixes that by listening for the close event and reopening it.

See https://github.com/godaddy/kubernetes-client/issues/520

## Why

The streams were closing prematurely on CRs which prevented the MDC from updating to reflect the actual state.

## Verification Steps
 
1. Run MDC.
2. Create an app. Bind the Push Notification service and bind both variants.
3. Leave MDC for 2-3 minutes and try to unbind again.
4. `mobile-services.json` should update.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -17,7 +17,10 @@ const KEYCLOAK_SECRET_SUFFIX = '-install-config';
 const DATASYNC_CONFIGMAP_SUFFIX = '-data-sync-binding';
 const MOBILE_SECURITY_SUFFIX = '-security';
 
+let updating = false;
+
 function updateAll(namespace, kubeclient) {
+  updating = true;
   console.log('Check services for all apps');
   return kubeclient.apis[mobileClientCRD.spec.group].v1alpha1
     .namespace(namespace)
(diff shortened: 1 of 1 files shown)


server/appServices.js
```diff
@@ -100,27 +97,86 @@ function updateAppsAndWatch(namespace, kubeclient) {
...
+    if (event.type === 'ADDED') {
+      updateAll(namespace, kubeclient);
+    }
+  });
+
+  // === WORKAROUND ===
+  // The stream automatically closes when watching CRDs after an inconsistent amount of time.
+  // Listen for when the stream ends and reopen it.
+  // Ref: https://github.com/godaddy/kubernetes-client/issues/520
+  appStream.on('end', () => {
```
wei-lee: actually, is there an `error` event that will be emitted, or does it always emit the `end` event?

craicoverflow: https://www.npmjs.com/package/JSONStream

There doesn't seem to be.

PR #350, opened 2019-07-31
Pf4 updates

dlabaj:
## Motivation
Added PF4 to mobile developer console and updated the Empty State Component

## Verification Steps
Verified that the Empty state was updated and the tests passed.

Screen Shot 2019-07-23 at 3 37 14 PM

## Checklist:

## Progress

- [x] Finished task

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@patternfly/patternfly": "^2.23.0",
+    "@patternfly/react-core": "^3.58.1",
+    "@patternfly/react-icons": "^3.10.13",
     "axios": "0.19.0",
     "body-parser": "1.19.0",
     "classnames": "2.2.6",
--- a/src/components/overview/MobileClientCardView.js
+++ b/src/components/overview/MobileClientCardView.js
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
-import { Toolbar, Filter, EmptyState, EmptyStateTitle, EmptyStateAction, CardGrid } from 'patternfly-react';
+import { Toolbar, Filter, CardGrid } from 'patternfly-react';
+import { EmptyState, EmptyStateBody, EmptyStateVariant, EmptyStateIcon, Title } from '@patternfly/react-core';
+import { MobileAltIcon } from '@patternfly/react-icons';
 import DebounceInput from 'react-debounce-input';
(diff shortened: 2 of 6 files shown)


src/components/overview/MobileClientCardView.js
```diff
@@ -11,7 +13,8 @@ class MobileClientCardView extends Component {
     this.state = { filter: '', currentValue: '' };
 
     this.emptyStateMessage = {
-      noAppsCreated: 'You have no mobile apps right now. Create one to get started.',
+      noAppsCreatedTitle: "You don't have any Mobile Apps.",
```
craicoverflow: Our [eslint configuration](https://github.com/aerogear/mobile-developer-console/blob/c151bb4532b63358224f41bee47a4ef78aa55850/.eslintrc.json) is set to prefer single quotes for JS files. I'm not sure why, but eslint is not actually picking up that specific line for some reason.

See below, I changed the next property in the object which gives an error, but not in `noAppsCreatedTitle`.

Screenshot from 2019-08-01 10-08-34

package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@patternfly/patternfly": "^2.23.0",
```
craicoverflow: Is it better to fix the version of this?

```suggestion
"@patternfly/patternfly": "2.23.0"
```

package.json
```diff
@@ -3,6 +3,9 @@
...
+    "@patternfly/react-core": "^3.58.1",
```
craicoverflow: Is it better to fix the version of this?

```suggestion
"@patternfly/react-core": "3.58.1"
```

package.json
```diff
@@ -3,6 +3,9 @@
...
+    "@patternfly/react-icons": "^3.10.13",
```
craicoverflow: Is it better to fix the version of this?

```suggestion
"@patternfly/react-icons": "3.10.13"
```

src/components/overview/MobileClientCardView.js
dlabaj: I'll update these to single quotes and fix the button.

package.json
dlabaj: yep all these should be fixed.  Will update these thanks

dlabaj:
Canceling PR we are going to do another PR with PF4 and Masthead changes.

PR #353, opened 2019-08-02
maxlength realmId

odra:
## Motivation
https://issues.jboss.org/browse/AGMS-619

## What
* adds field value size limit for the auth realmId field (36)

## Why
Keycloak database realm id/name schema has a max size limit of 36 chars.

## How
* Created a new MaxLengthRule validation rule
* RealmId field can now be edited when binding an auth service

## Verification Steps
* Build an image using this branch and deploy it
* Bind an auth service using a realm id that has a length greater than 36 chars 
* The form should throw an error
* Edit the realm id value and use a value that is small enough
* Form submission should work now.
## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

I have an image available with the mentioned changes: `docker.io/odranoel/mdc:dev`

### DIFF
--- a/src/components/mobileservices/validator/rules/MaxLengthRule.js
+++ b/src/components/mobileservices/validator/rules/MaxLengthRule.js
@@ -1,0 +1,34 @@
+import { get } from 'lodash-es';
+import { ValidationRuleBaseClass } from './ValidationRuleBaseClass';
+
+/**
+ * This is a very simple rule. Checks that the configured field value length is greater than the one that is expected.
+ * It must be configured inside the JSON file as follows:
+ * {
+ *   "type": "maxlength",  // this must be exactly 'maxlength'
+ *   "length" 36,          // the maximum length limit of this field value
+ *   "error": "Your error message"          // Optional. The custom error message.
+ *                                          // This can be a string, a dictionary or a function (signature: (key, message) => {})
+ *                                          // If not specified, a standard error message is returned.
+ * }
+ */
+export class MaxLengthRule extends ValidationRuleBaseClass {
+  constructor(config) {
+    super(config);
+    this.maxlength = config.length;
+  }
+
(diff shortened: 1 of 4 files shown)


src/models/mobileservices/keycloakrealmcr.js
```diff
@@ -90,6 +87,15 @@ export class KeycloakRealmCR extends CustomResource {
           comment: 'This is the set of rules that will be used to validate IDM bindings',
           fields: {
             realmSettings: {
+              realmId: {
+                validation_rules: [
+                  {
+                    type: 'maxlength',
+                    length: 36,
+                    error: 'realm id is too big, max length is 36 chars'
```
craicoverflow: I would suggest:

"Realm ID is too long, maximum length is 36 characters".

src/components/mobileservices/validator/rules/MaxLengthRule.js
```diff
@@ -0,0 +1,34 @@
+import { get } from 'lodash-es';
+import { ValidationRuleBaseClass } from './ValidationRuleBaseClass';
+
+/**
+ * This is a very simple rule. Checks that the configured field value length is greater than the one that is expected.
```
ziccardi: The description here is a bit confusing.
I think it would better 'Checks that the passed in value has a length smaller or equal than the configured maxlenght'

src/models/mobileservices/keycloakrealmcr.js
ziccardi: There is no point in hardcoding the message here. Better to remove the error message here and just use the default one (see previous comment)

a
```diff
@@ -0,0 +1,261 @@
+
```
grdryn: @odra This looks like it was committed by accident, was it?

PR #354, opened 2019-08-02
fix(mobile-services): re-open streams after they automatically close

craicoverflow:
https://issues.jboss.org/browse/AEROGEAR-9668

`mobile-services.json` will stop updating after a number of minutes (30+). This is because of the `--min-request-timeout` flag in kube-apiserver. See https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/.

This fix reopens the watch endpoint on each of the mobile services when the stream detects an endpoint getting closed.

## Verification

This is hard to verify since the Mobile Developer Console needs to be running uninterrupted for a long time before the watches close. I suggest you follow the steps below, but also leave it for another 30 minutes afterwards to see if it is still working.

1. Ensure MDC is running for a long time (> 40 minutes to be sure).
2. Create an app and try to bind Mobile Security.
3. Check `mobile-services.json`. It should be updated.

### DIFF
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -68,35 +68,17 @@ function getServicesForApp(namespace, app, kubeclient) {
 async function updateAppsAndWatch(namespace, kubeclient) {
   updateAll(namespace, kubeclient).then(async () => {
     watchMobileClients(namespace, kubeclient);
-
-    const secretStream = await kubeclient.api.v1.watch.namespace(namespace).secrets.getObjectStream();
-    secretStream.on('data', event => {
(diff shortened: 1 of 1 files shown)


Boomatang:
I may have an issue with my set up but when I do the validation steps I first did a control run which fails.
1. Start MDC and login. (not letting MDC run for a long time)
2. Create an app and try to bind Mobile Security. The mobile security is running locally using the latest Mobile Security Service Operator
3. Check `mobile-services.json`. It should be updated. **There was no change to this file in the MDC on direct start up.**

PR #355, opened 2019-08-07
feat(validation): prevent mobile security service binding if app id exists

craicoverflow:
https://issues.jboss.org/browse/AGMS-701

This change will prevent multiple apps with the same unique ID getting bound to the Mobile Security Service.

1. Create an app.
2. Bind Mobile Security with an app ID "com.myapp".
3. Create a second app.
4. Try to bind Mobile Security again with the same app ID.
5. You should get an error message and the binding will be prevented.


ziccardi:
👀

PR #356, opened 2019-08-07
Change Mobile Client CRD api group

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9691


aliok:
I see the version in package.json is not updated since long time, so I didn't touch it.
cc @craicoverflow

deploy/mobile-client-crd.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: mobileclients.mobile.k8s.io
+  name: mobileclients.mobile.mdc.aerogear.org
```
craicoverflow: This value should be `mobileclients.mdc.aerogear.org`. Otherwise the following error message is given when trying to create the CRD:

_The CustomResourceDefinition "mobileclients.mobile.mdc.aerogear.org" is invalid: metadata.name: Invalid value: "mobileclients.mobile.mdc.aerogear.org": must be spec.names.pl
ural+"."+spec.group"._

grdryn: Looking at the rest of this PR, I think the `.mobile` part should also be removed here.

i.e.

```
mobileclients.mdc.aerogear.org
```
rather than

```
mobileclients.mobile.mdc.aerogear.org
```

aliok: Thanks for the catch @craicoverflow @grdryn 
Pushed a new commit already

PR #364, opened 2019-08-08
Fixed Security issues

ziccardi:
## Motivation
Jira:

- https://issues.jboss.org/browse/AGMS-649
- https://issues.jboss.org/browse/AGMS-650
- https://issues.jboss.org/browse/AGMS-651


camilamacedo86:
doing the review.

src/components/build/BuildInformation.test.js
```diff
@@ -55,11 +55,11 @@ describe('BuildInformation', () => {
...
   });
 
-  describe('BuildDownload', () => {
-    it('should match previous snapshot', () => {
-      const BuildDownload = wrapper.find('BuildDownload');
-
-      expect(BuildDownload).toMatchSnapshot();
-    });
-  });
+  // describe('BuildDownload', () => {
```
craicoverflow: I think this block should be completely removed instead of just commented out.

camilamacedo86: 👍

ziccardi: The idea was to leave it there to be reintroduced in the future. However, since that page is not used anymore, probably it would be ok to remove it

PR #377, opened 2019-08-14
refactor: make anonymous function synchronous and remove unnecessary condition.

craicoverflow:
- Made function synchronous as it no longer needed to be async.
- Remove condition which prevented mobileclients from getting updated multiple times in parallel. The apps will still update fine without it.

## Verification

Binding/unbinding should still work as normal. You will see 409 errors in the console but this is okay as MDC tries to update the app multiple times.


craicoverflow:
> I have issue with the stack trace been printed for the 409. this should be handled in a different manner. With that said the most of the logging does require a refactor, so it might be better to fix it at that point.

I agree, the current logging can do with a lot of improvements which we should definitely tackle in the next stage.

Issue #378, opened 2019-08-14
Fix styling issues


christiemolloy:
## Description
Fix the background for pages from grey to white
fix some spacing issues.

## Environment
* Operating system:
* OpenShift versions:
* Project Versions: 

## Steps to reproduce
Go to the "mobile client" page

![Screenshot from 2019-08-14 12-54-56](https://user-images.githubusercontent.com/20118816/63053652-4ef96700-beb0-11e9-98a0-f07f5a6f6b8f.png)
![Screenshot from 2019-08-14 12-59-37](https://user-images.githubusercontent.com/20118816/63053653-4ef96700-beb0-11e9-8526-e36690d02e83.png)

christiemolloy:
closing this because PR was merged

PR #379, opened 2019-08-14
fix(pages): update styles

christiemolloy:
## Motivation
https://github.com/aerogear/mobile-developer-console/issues/378

## What
Fixed styles after the new patternfly updates

## Why
the pf updates didn't account for some of the styles

## How
Add PageSectionVariants to the app.js file and change it to light.

## Verification Steps
I added screenshots, but you could run this locally and check the background color
 
## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

Screen Shot 2019-08-14 at 4 22 10 PM

Screen Shot 2019-08-14 at 4 22 23 PM

I was also seeing this when I ran the code so unset the width:
Screen Shot 2019-08-14 at 4 43 15 PM

### DIFF
--- a/src/components/configuration/ConfigurationView.css
+++ b/src/components/configuration/ConfigurationView.css
@@ -6,10 +6,15 @@
     width: 30px;
     height: 30px; }
   .configurationView .list-group-item-container {
-    padding-top: 24px;
-    padding-right: 36px;
-    padding-bottom: 24px;
-    padding-left: 36px; }
+    padding-top: 16px;
+    padding-right: 24px;
+    padding-bottom: 16px;
+    padding-left: 24px; }
+    .configurationView .list-group-item-container ol {
+      margin-left: 32px; }
+    .configurationView .list-group-item-container ol > li > ul {
+      list-style-type: disc;
+      margin-left: 32px; }
   .configurationView .list-group-item-header {
     background-color: transparent;
     margin-top: 8px;
--- a/src/components/configuration/ConfigurationView.css
+++ b/src/components/configuration/ConfigurationView.css
@@ -25,7 +25,6 @@
     margin-right: 8px; }
 
 .configurationView > * {
-  flex: 1 100%;
   margin-right: 16px;
   display: flex;
   flex-direction: column; }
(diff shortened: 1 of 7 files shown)


src/components/configuration/ConfigurationView.css
```diff
@@ -5,6 +5,11 @@
   .configurationView .icon {
     width: 30px;
     height: 30px; }
+  .configurationView .list-group-item-container {
```
christiemolloy: @dlabaj for some reason this isn't in the codebase but I think its rendered some how. Whats the best way to remove it?

craicoverflow: If you run `npm run build-css` it should remove this block from the file.

craicoverflow:
Could you also request review from @ziccardi on future PRs to this project? He is the other engineer working on this so you have a better chance of one or the other of us picking it up.

src/components/configuration/ConfigurationView.css
christiemolloy: @craicoverflow yeah I only added it to the `.scss` file, but when I run `npm run build-css` it removes it from the `css`... thanks for that

christiemolloy:
@craicoverflow thanks the reviews. I'm currently seeing that the code block is not overflowing, and I have the ability to scroll. I provided a screenshot underneath to show you what I am seeing.

Screen Shot 2019-08-16 at 11 31 34 AM

Also fixed the list styling here, let me know if theres any more updates you would like to see here:

Screen Shot 2019-08-16 at 11 32 11 AM

Also, when you mentioned that the title is overflowing out of the container, what would you like to see here? I would need a design to implement the desired behavior.

craicoverflow:
Apologies, I completely missed this.

So the scrolling works fine in Chrome, but not in Firefox and it is happening on master so ignore that point.

On the second point, currently on master there seems to be a minimum width, so the browser window never gets small enough and no overflow can occur.

Screenshot from 2019-08-22 16-19-44

craicoverflow:
> @craicoverflow looks like `flex: 1 100%` was being set on the copy to clipboard component which prevented it from resizing in its parent container. Now both columns stay at 50% width and the clipboard scrolls. I added a screenshot below. Let me know if this addressed everything you were seeing.
> 
> Screen Shot 2019-08-26 at 10 04 16 AM

Looks good, thanks!

PR #380, opened 2019-08-15
fix: enable server to retry connecting to kube api

craicoverflow:
This fixes an issue when Kubernetes API restarts while the Node server is running, which causes the atches on the resources to fail. 

This retries the watches for n (default 10) times before exiting the application.

1. Run the MDC server against your remote cluster.
2. SSH into the master node.
3. Restart the API and controllers:

```sh
/usr/local/bin/master-restart api
```

and

```sh
/usr/local/bin/master-restart controllers
```

Watch the logs from MDC. You should see the server failing to attempting to set up the watch on the resources again, eg:

```
waiting 3000 milliseconds before attempting services watches again
```

Once it has successfully reconnected you should see `connection successful`.

4. Refresh MDC and try to bind a service. It should be successful.

### DIFF
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -3,6 +3,7 @@ const mobileClientCRD = require('./mobile-client-crd.json');
 const androidVariantCRD = require('./android-variant-crd.json');
 const iosVariantCRD = require('./ios-variant-crd.json');
 const mobileSecurityServiceCRD = require('./mobile-security-crd.json');
+const { debounce } = require('lodash');
 const {
   PushService,
   IdentityManagementService,
@@ -17,7 +18,18 @@ const KEYCLOAK_SECRET_SUFFIX = '-install-config';
 const DATASYNC_CONFIGMAP_SUFFIX = '-data-sync-binding';
 const MOBILE_SECURITY_SUFFIX = '-security';
 
+const { MAX_RETRIES = 60, RETRY_TIMEOUT = 5000 } = process.env;
+
 let updating = false;
+// keeps track of how many attempts were made to connect to the Kubernetes API server
+let connectionAttempts = 0;
+const debounceTime = 250;
+// create a debounce function for the if a connection to
+// the kubernetes API fails. this will ensure that only one connection
+// retry attempt will take place at a time.
+const retryWatchDebounce = debounce((namespace, kubeclient) => {
+  retryWatch(namespace, kubeclient);
+}, debounceTime);
 
 function updateAll(namespace, kubeclient) {
   updating = true;
@@ -32,7 +44,13 @@ function updateAll(namespace, kubeclient) {
       updating = false;
       return Promise.all(promises);
     })
-    .catch(err => console.error(`Failed to update apps due to error`, err));
+    .catch(err => {
+      console.error(`Failed to update apps due to error`, err);
+
+      retryWatchDebounce(namespace, kubeclient);
+
+      throw err;
+    });
 }
 
 function updateApp(namespace, app, kubeclient) {
@@ -57,7 +75,13 @@ function updateApp(namespace, app, kubeclient) {
       }
       return Promise.resolve();
     })
-    .catch(err => console.error(`Failed to update app ${app.metadata.name} due to error`, err));
+    .catch(err => {
+      console.error(`Failed to update app ${app.metadata.name} due to error`, err);
+
+      retryWatchDebounce();
+
+      throw err;
+    });
 }
 
 function getServicesForApp(namespace, app, kubeclient) {
@@ -67,13 +91,16 @@ function getServicesForApp(namespace, app, kubeclient) {
 }
 
 async function updateAppsAndWatch(namespace, kubeclient) {
-  updateAll(namespace, kubeclient).then(async () => {
+  connectionAttempts++;
+  updateAll(namespace, kubeclient).then(() => {
     watchMobileClients(namespace, kubeclient);
     watchDataSyncConfigMaps(namespace, kubeclient);
     watchKeyCloakSecrets(namespace, kubeclient);
     watchAndroidVariants(namespace, kubeclient);
     watchIosVariants(namespace, kubeclient);
     watchMobileSecurityApps(namespace, kubeclient);
+    // reset connection attempts count
+    connectionAttempts = 1;
   });
 }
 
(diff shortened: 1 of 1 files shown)


server/appServices.js
```diff
@@ -138,6 +144,22 @@ async function watchKeyCloakSecrets(namespace, kubeclient) {
...
+    retryWatch(namespace, kubeclient);
+  });
+}
+
+function retryWatch(namespace, kubeclient) {
+  if (connectAttempts < MAX_RETRIES) {
+    console.log(`waiting ${RETRY_TIMEOUT} milliseconds before attempting services watches again`);
+    setTimeout(() => {
+      updateAppsAndWatch(namespace, kubeclient);
+    }, 30000);
```
ziccardi: This should be RETRY_TIMEOUT

server/appServices.js
```diff
@@ -138,6 +144,22 @@ async function watchKeyCloakSecrets(namespace, kubeclient) {
...
   });
+
+  secretStream.on('close', () => {
+    retryWatch(namespace, kubeclient);
+  });
+}
+
+function retryWatch(namespace, kubeclient) {
+  if (connectAttempts < MAX_RETRIES) {
+    console.log(`waiting ${RETRY_TIMEOUT} milliseconds before attempting services watches again`);
```
ziccardi: I would suggest to add here an info like 'ATTEMPT X of Y: ....'

server/appServices.js
```diff
@@ -138,6 +144,22 @@ async function watchKeyCloakSecrets(namespace, kubeclient) {
   secretStream.on('end', () => {
     watchKeyCloakSecrets(namespace, kubeclient);
```
ziccardi: Why keeping this?

server/appServices.js
```diff
@@ -138,6 +144,22 @@ async function watchKeyCloakSecrets(namespace, kubeclient) {
...
   });
+
+  secretStream.on('close', () => {
+    retryWatch(namespace, kubeclient);
```
ziccardi: Why do we do the retry only when the keycloak stream is closed? 
Can we take for granted that it will never happen that, for example, the datasync stream is closed while the keycloak one is not?

craicoverflow: If the Kubernetes API is restarted then we will 100% not be able to connect to *any* resource. If we do this on every resource then we would be trying to reconnect to the server multiple times in parallel which is not needed.

ziccardi: Yes, but are we sure that API restart is the only event that could cause the stream to be closed?

server/appServices.js
craicoverflow: `end` is different. It is when the watch on the resource is closed. `close` happens when the kubernetes API is restarted.

```

--min-request-timeout int     Default: 1800
An optional field indicating the minimum number of seconds a handler must  keep a request open before timing it out. Currently only honored by the  watch request handler, which picks a randomized value above this number  as the connection timeout, to spread out load.

```

ziccardi: Are you sure this doesn't result in a `close` event too?

```
The 'end' event is emitted when there is no more data to be consumed from the stream.
```

server/appServices.js
wei-lee: maybe not, but I don't think it really matters, because when reconnect, it will reconnect for all resources. The only danger here is that is it possible that for whatever reason, all the other connections are closed but the secret one is not.  I am not sure if there is such a case, but if this is a concern, a safer way could be that we also listen to the `close` event on all the resources, but in the function `retryWatch` function itself,  ignore some of them if they are emitted in a very short of time period (something like a `debounce` function).

ziccardi: @wei-lee 👍  That was exactly my concern. We could have any stream closed, not just keycloak (not only all but keycloak, but any but keycloak)

server/appServices.js
```diff
@@ -138,6 +145,33 @@ async function watchKeyCloakSecrets(namespace, kubeclient) {
...
+  console.log('connection to Kubernetes API lost');
+
+  if (connectAttempts < MAX_RETRIES) {
+    console.log(
+      `waiting ${RETRY_TIMEOUT} milliseconds before attempting services watches again`,
+      `attempt ${connectAttempts} of ${MAX_RETRIES}`
+    );
+    setTimeout(() => {
+      updateAppsAndWatch(namespace, kubeclient);
+    }, RETRY_TIMEOUT);
```
wei-lee: this will only try the `updateAppsAndWatch` once, what happens if the api server is not back yet and the connection failed? Will it actually try again?

ziccardi: Maybe something like [this](https://github.com/aerogear/mobile-developer-console/blob/688c7648ee09a49ae31f7b5a165d7155b9532dd9/src/services/crmanagers/PushVariantResourceManager.js#L5) would help here

server/appServices.js
```diff
@@ -30,9 +46,16 @@ function updateAll(namespace, kubeclient) {
...
     .then(promises => {
       updating = false;
+      connectionAttempts = 0;
       return Promise.all(promises);
     })
-    .catch(err => console.error(`Failed to update apps due to error`, err));
+    .catch(err => {
+      console.error(`Failed to update apps due to error`, err);
+
+      if (err.code === ECONNREFUSED) {
```
ziccardi: If the error is not `ECONNREFUSED`, the stream is closed, but the app doesn't exits, since the `retryWatch` (containing the exit) is never called: in this case the MDC won't be restarted, but the `mobile-services.json` update will never work

server/appServices.js
```diff
@@ -17,9 +18,24 @@ const KEYCLOAK_SECRET_SUFFIX = '-install-config';
 const DATASYNC_CONFIGMAP_SUFFIX = '-data-sync-binding';
 const MOBILE_SECURITY_SUFFIX = '-security';
 
+const { MAX_RETRIES = 360, RETRY_TIMEOUT = 5000 } = process.env;
```
ziccardi: This will make the MDC retry for 30 minutes. Shouldn't be about 5 minutes instead?

craicoverflow: you're right. thanks!

server/appServices.js
craicoverflow: I see your point.

We know that `ERRCONNREFUSED` will close the streams (as tested). Do you suggest we do not add a condition here and call `retryWatch` whenever any error is caught here instead? 

My concern is that this might set up extra watches again when they are already open. One thing we could do is set a flag `isKubeApiDown=true` in `.x.on('close',` and instead of checking `ERRCONNREFUSED`, check if `isKubeApiDown` instead. What do you think?

PR #382, opened 2019-08-19
refactor: added max length validation

craicoverflow:
https://issues.jboss.org/browse/AEROGEAR-9768

Added max length validation to the app name.

## Verification

1. Try to create an app > 70 characters. It should be prevented.

### DIFF
--- a/src/components/create_client/EditMobileClientBaseClass.js
+++ b/src/components/create_client/EditMobileClientBaseClass.js
@@ -1,13 +1,13 @@
 /* eslint guard-for-in: 0 */
 import React, { Component } from 'react';
 import { Grid, Form, Row, Col } from 'patternfly-react';
-import { MobileApp } from '../../models';
+import { MobileApp, MAXLENGTH_APPNAME } from '../../models';
 import { CREATE_CLIENT_NAME } from './Constants';
 import { VerticalFormField } from './VerticalFormField';
 
 export const LABEL_APPNAME = '* App Name';
 export const EXAMPLE_APPNAME = 'myapp';
-export const HELP_APPNAME = 'App name must only contain lowercase letters, numbers and dots.';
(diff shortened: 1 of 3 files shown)


src/models/mobileapps/mobileapp.js
```diff
@@ -43,8 +43,12 @@ export default class MobileApp {
     const value = this.getProperty(propertyName);
     if (value) {
       switch (propertyName) {
-        case PROPERTIES.NAME:
+        case PROPERTIES.NAME: {
+          if (value.length > 100) {
```
ziccardi: Better to make this a constants instead than hardcoding it here

src/components/create_client/EditMobileClientBaseClass.js
```diff
@@ -7,7 +7,8 @@ import { VerticalFormField } from './VerticalFormField';
 
 export const LABEL_APPNAME = '* App Name';
 export const EXAMPLE_APPNAME = 'myapp';
-export const HELP_APPNAME = 'App name must only contain lowercase letters, numbers and dots.';
+export const HELP_APPNAME =
+  'App name can not exceed 100 characters, and must only contain lowercase letters, numbers and dots.';
```
ziccardi: Better to make `100` a constants instead than hardcoding it here

craicoverflow:
> Some points I have noted
> 
>     * No service will bind with more than 63 characters for the app name.
> 
>     * The name does not display correctly on the landing page cards if longer than 75 characters

These are pre-existing problems though, right?

Boomatang:
The issue with the name been more 63 characters has been before with labels in the CR's. But the 75+ characters long name not showing correctly is only some thing I noticed today. There is no other issue created for it.

craicoverflow:
@ziccardi I have changed the max length to be 70 characters. I believe this is more than sufficient. We could reduce it to be 50 to be more like Android and iOS, but I don't believe we need to. What do you think?

PR #385, opened 2019-08-20
Fixed issue AEROGEAR-8283

ziccardi:
## Motivation

JIRA: https://issues.jboss.org/browse/AEROGEAR-8283 

## What
Modified the `mobileservice.isBindingOperationInProgress` so that it can receive an `owner` parameter. This way we will be able to ask if a binding operation is in progress for a particular application

Modified the BindingPanel to get closed automatically only if the wizard has reached the latest step and the binding was in progress. This is to avoid that another binding for the same app could cause this window to get closed before the latest step has been reached (since `isBindingOperationInProgress(appName)` would return `false` since this operation has not started yet).

## Why
The binding panel in MDC gets automatically closed when `isBindingOperation` returns 'false'.
Without this change, any binding operation on any application will cause all the opened Binding Panels to get closed, since the `isBindingOperation` is not filtered for a particular application.

## How
Added an `owner` parameter to the `isBindingOperation` method and checking if the wizard has reached the latest step if a `close binding panel` event is received.

## Verification Steps
 
**CASE1**

1. Connect to the mobile-console by using 2 different browsers (using a chrome instance and a chrome incognito mode will work)
2. Create 2 applications (testapp1 and testapp2)
3. Go to the first browser and open testapp1, then start a binding (don't finish it)
4. Go to the second browser and open testapp2, the complete a binding
5. Without the fix, the window on the first browser would close.

**CASE2**

1. Connect to the mobile-console by using 2 different browsers (using a chrome instance and a chrome incognito mode will work)
3. Go to the first browser and open testapp1, then start a binding (don't finish it)
4. Go to the second browser and open testapp1 again, then complete a binding
5. Without the fix, the window on the first browser would close.

## Checklist:

## Known issues
The window will close automatically when **ALL THE BINDINGS** for the current application and for the selected service will finish. I.E. If an Android and an iOS binding are in progress, the window will automatically close when both of them will end. 
**This is not a regression**: without the fix the window would have closed when all the bindings of the given type would have ended (regardless of the application).

### DIFF
--- a/src/components/mobileservices/BindingPanel.js
+++ b/src/components/mobileservices/BindingPanel.js
@@ -138,6 +138,10 @@ export class BindingPanel extends Component {
     }
   };
 
+  isInProgress() {
+    return this.state.loading;
+  }
+
   /**
    * see https://github.com/mozilla-services/react-jsonschema-form/tree/6cb26d17c0206b610b130729db930d5906d3fdd3#form-data-validation
    */
@@ -203,5 +207,7 @@ function mapStateToProps(state, ownProps) {
 
 export default connect(
   mapStateToProps,
-  mapDispatchToProps
+  mapDispatchToProps,
+  null,
+  { withRef: true }
 )(BindingPanel);
--- a/src/components/mobileservices/BindingStatus.js
+++ b/src/components/mobileservices/BindingStatus.js
@@ -11,12 +11,12 @@ class BindingStatus extends Component {
   }
 
   updateState() {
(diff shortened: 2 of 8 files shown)


ziccardi:
@craicoverflow I believe it should not close (and it doesn't close with this fix): when he will complete the binding, he will get an appropriate error.
This is especially true for the PUSH SERVICE: if user A opens the BindingPanel to create an Android Binding and user B opens the BindingPanel to create an iOS binding, the bound service is always the same, but user A will be able to correctly finish the binding process.

PR #387, opened 2019-08-20
Feat(ErrorMessages): Added better error message around creating new apps that already exist

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9711

## What
Add response checking for the ErrorMessage component to handle kind of MobileClient CR

## Why
To add better messages displayed to the users

## How
In ErrorMessage.js there is checks in place looking for the `response.data.details.kind` of `mobileclients` and `response.data.reason` of `AlreadyExists` is the error type. This should only happen when a new app is been created with an existing name.

## Verification Steps
1. Create a new app, eg: myapp
2. Try create a new app with the existing app name, myapp
3. Error message shown will display a meaningfully error message
4. Error message should close when X is pressed

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

image

### DIFF
--- a/src/containers/ErrorMessages.js
+++ b/src/containers/ErrorMessages.js
@@ -2,7 +2,43 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import { ToastNotificationList, ToastNotification } from 'patternfly-react';
 import { withRouter } from 'react-router-dom';
+import { get } from 'lodash-es';
 import { dismiss, dismissAll } from '../actions/errors';
+
+const MOBILECLIENT = 'mobileclients';
+const ALREADYEXISTS = 'AlreadyExists';
+
+function errorMessage(error) {
+  try {
+    switch (error.response.data.details.kind) {
+      case MOBILECLIENT:
+        // case to handle MobileClient CRs
+        return { displayMessage: mobileClientError(error), message: error.message };
+      default:
+        return { message: error.message };
+    }
+  } catch (TypeError) {
+    // catch any error if error.response.data.details.kind does not exist
+    return { message: error.message };
+  }
+}
+
+function mobileClientError(error) {
+  try {
+    switch (error.response.data.reason) {
+      case ALREADYEXISTS: {
+        // handle if mobileClient exists
+        const appName = get(error, 'response.data.details.name');
+        return `An app named "${appName}" already exists`;
+      }
+      default:
+        return error.message;
+    }
+  } catch (TypeError) {
+    // catch any errors if error.response.data.reason does not exist
+    return error.message;
+  }
+}
 
 export class ErrorMessages extends Component {
   componentWillMount() {
@@ -13,14 +49,13 @@ export class ErrorMessages extends Component {
   componentWillUnmount() {
     this.unlisten();
   }
-
   render() {
     const { errors, dismissError } = this.props;
     return (
       <ToastNotificationList>
-        {[...new Set(errors.map(error => error.message))].map((error, index) => (
-          <ToastNotification key={index} onDismiss={() => dismissError(error)}>
-            <span>{error}</span>
+        {[...new Set(errors.map(error => errorMessage(error)))].map((error, index) => (
+          <ToastNotification key={index} onDismiss={() => dismissError(error.message)}>
+            <span>{error.displayMessage || error.message}</span>
           </ToastNotification>
         ))}
       </ToastNotificationList>
@@ -43,3 +78,5 @@ export default withRouter(
     mapDispatchToProps
   )(ErrorMessages)
 );
+
+export { errorMessage, mobileClientError, MOBILECLIENT, ALREADYEXISTS };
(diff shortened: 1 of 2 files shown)


src/containers/ErrorMessages.js
```diff
@@ -13,14 +46,13 @@ export class ErrorMessages extends Component {
...
   render() {
     const { errors, dismissError } = this.props;
     return (
       <ToastNotificationList>
-        {[...new Set(errors.map(error => error.message))].map((error, index) => (
-          <ToastNotification key={index} onDismiss={() => dismissError(error)}>
-            <span>{error}</span>
+        {[...new Set(errors.map(error => errorMessage(error)))].map((error, index) => (
+          <ToastNotification key={index} onDismiss={() => dismissError(error.message)}>
+            <span>{error.dispalyMessage || error.message}</span>
```
ziccardi: I guess this should be `displayMessage`

src/containers/ErrorMessages.js
```diff
@@ -4,6 +4,39 @@ import { ToastNotificationList, ToastNotification } from 'patternfly-react';
...
+    return { message: error.message };
+  }
+}
+
+function mobileClientError(error) {
+  try {
+    switch (error.response.data.reason) {
+      case ALREADYEXISTS:
+        // handle if mobileClient exists
+        return 'App name currently exists';
```
ziccardi: Do you have the app name here? If you have, I think that an error like `An app named ${app.name} already exists`.
If we don't have the app name here, I believe the message would be better as 'An app with the same name already exists`

src/containers/ErrorMessages.test.js
```diff
@@ -25,3 +25,49 @@ describe('ErrorMessages', () => {
...
+  it('value response.data.reason does not exist', () => {
+    const expected = 'error message';
+    const error = { message: 'error message' };
+    expect(mobileClientError(error)).toEqual(expected);
+  });
+});
+
+describe('errorMessage() returns correct value', () => {
+  it('error message for a mobileClient CR Switch Case', () => {
+    const expected = { dispalyMessage: 'App name currently exists', message: 'error message' };
```
ziccardi: I believe this should be `displayMessage`

src/containers/ErrorMessages.test.js
```diff
@@ -25,3 +25,49 @@ describe('ErrorMessages', () => {
     ).toEqual('error2');
   });
 });
+
+describe('mobileClientError() returns correct value', () => {
+  it('return alreadyesixts Switch Case error message', () => {
+    const expected = 'App name currently exists';
+    const error = { message: 'error message', response: { data: { reason: ALREADYEXISTS } } };
```
ziccardi: Would be better to create a constant with the `message`

src/containers/ErrorMessages.js
```diff
@@ -27,7 +27,7 @@ function mobileClientError(error) {
     switch (error.response.data.reason) {
       case ALREADYEXISTS:
         // handle if mobileClient exists
-        return 'App name currently exists';
+        return `An app named "${error.response.data.details.name}" already exists`;
```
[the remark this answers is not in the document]
ziccardi: Yeah, better to use `get` from `lodash-es`

src/containers/ErrorMessages.js
```diff
@@ -24,10 +25,11 @@ function errorMessage(error) {
 
 function mobileClientError(error) {
   try {
+    const appName = get(error, 'response.data.details.name');
```
craicoverflow: I would move this line inside the switch case as it is not needed until then.

src/containers/ErrorMessages.test.js
```diff
@@ -50,6 +50,15 @@ describe('mobileClientError() returns correct value', () => {
     const error = { message: ERRORMESSAGE };
     expect(mobileClientError(error)).toEqual(expected);
   });
+
+  it('value resonse.data.details.name does not exist', () => {
```
craicoverflow: resonse should be response.

src/containers/ErrorMessages.js
craicoverflow: Typo: should be "displayMessage".

src/containers/ErrorMessages.js
Boomatang: If this is move inside the switch case, there is errors raised by `eslint`. Should I still move it.

craicoverflow: Once you add curly brackets to your switch case, these warnings will go away.

src/containers/ErrorMessages.js
Boomatang: This has been fixed in a different commint

src/containers/ErrorMessages.test.js
Boomatang: Will fix

src/containers/ErrorMessages.js
Boomatang: :+1:

PR #388, opened 2019-08-22
refactor: compare app label value against app name and UID

craicoverflow:
https://issues.jboss.org/browse/AEROGEAR-9779

Follow up to https://github.com/aerogear/mobile-developer-console/pull/386.

This change will make the change introduced in https://github.com/aerogear/mobile-developer-console/pull/386 backwards compatible, by checking the app label for the new value type (app UID) as well as the old value type (app name).

## Verification

1. Create a new app.
2. Bind all services.
3. The services should show in the correct lists (Bound Services or Unbound Services).
4. Inspect the service CRs. They should no longer have any labels.

### DIFF
--- a/src/models/mobileservices/mobileservice.js
+++ b/src/models/mobileservices/mobileservice.js
@@ -44,12 +44,12 @@ export class MobileService {
     return this.customResources.length > 0 && find(this.customResources, cr => cr.isReady());
   }
 
-  isBoundToApp(uid) {
-    return this.customResources.length > 0 && find(this.customResources, cr => cr.isReady() && cr.hasAppLabel(uid));
+  isBoundToApp(appName) {
+    return this.customResources.length > 0 && find(this.customResources, cr => cr.isReady() && cr.isOwner(appName));
   }
 
-  getCustomResourcesForApp(uid) {
-    return filter(this.customResources, cr => cr.hasAppLabel(uid));
+  getCustomResourcesForApp(appName) {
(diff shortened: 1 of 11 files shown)


ziccardi:
👀

src/models/mobileservices/mobileservice.js
```diff
@@ -64,14 +66,16 @@ export class MobileService {
...
-      this.customResources,
-      cr => (owner ? find(cr.getOwners(), o => o.name === owner) : true) && cr.isInProgress()
-    );
+  isBindingOperationInProgress(appName) {
+    const inprogressCR = find(this.customResources, cr => this.isOwner(cr, appName) && cr.isInProgress());
     return inprogressCR != null;
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  isOwner(cr, appName) {
```
ziccardi: I believe this logic should be part of the [customresource](https://github.com/aerogear/mobile-developer-console/blob/17fcb54ad98a6c051a6e6b88506bb0512f459344/src/models/mobileservices/customresource.js#L11) object by adding there, for example, an `isOwnedBy(appName)`.

That would bring 2 benefits:

1. the method will be automatically available to all the piece of codes that needs to deal with `customresources`
2. the method above could be changed to:
```javascript
const inprogressCR = find(this.customResources, cr => cr.isOwnedBy(appName) && cr.isInProgress());
```
making that more readable.

PR #394, opened 2019-08-28
fix(mobileClient): upgrade to pf4 components + new design

christiemolloy:
**WIP don't merge** 

## Motivation
No issue link. Goal to update the Mobile Client page to use PatternFly 4 components and to adhere with the new designs here https://marvelapp.com/4b4g0aa/screen/58518807

## What
All components were updated on the Mobile Client page and new designs are reflected.

## Why
This was done as part of the pf4 migration efforts.

## How
By replacing pf3 and mdc components with pf4 components.

## Verification Steps
1. Run a local build
2. Create a new app
3. Test the mobile client page

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

Screen Shot 2019-08-28 at 2 22 34 PM

Screen Shot 2019-08-28 at 2 31 10 PM


Boomatang:
I am aware this is a WIP but here is some of my comments.
1. On point 4. When was the decision to add back in the edit app name function taken. From what I am aware of this has big impact on the engineering side. Of course I will check this out with the rest of the team.
2. From a visual point of view I see dashed lines around buttons and drop-downs.
3. On the side panel where the "Full Mobile Config" s placed as imaged. There is a few points/questions I would like to get cleared up.
  a) The text description will want to be changed. The wording is misleading. That's not a big issue just pointing it out.
  b) It's understood that the mobile-services.json is not working right now. As a mock up how would that look when extended or collapsed.
  c) Where does the set up steps that were in the previous version go or have they been removed as part of the redesign. If they have been removed as part of the redesign, should these step be added somewhere. Note that the step where updated when a service is bound to the app.
4. Maybe the "Available Services" needs to push down a bit when there is bound services, see image. This seems to grand when there is no bound services.

For the most part I think its looking much better.

christiemolloy:
@boomatang thanks for your feedback this is the design that I was working from https://marvelapp.com/4b4g0aa/screen/58518810 so tagging @tiffanynolan to follow up to some of your design-related questions

PR #415, opened 2019-09-09
Agms 711

Boomatang:
## Motivation
https://issues.jboss.org/browse/AGMS-711

## What
Re-factor the server updated

## Why
Server event calls increase exponentially with every app added. This may be related to other issues with services not binding to MDC

## How
Add an short answer for: How it was done? (E.g By removing this feature from ... OR By removing just the button but not its implementation ... ) 

## Verification Steps
1. add apps to MDC
2. A reduced number of logs should be seen
3. A services to the apps
4. A reduced numbers of logs should be seen also the `mobile-services.json` should be updated with the new services.

* Try with a high number of apps 100+

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -168,8 +168,12 @@ async function watchMobileClients(namespace, kubeclient) {
 async function watchDataSyncConfigMaps(namespace, kubeclient) {
   const configmapStream = await kubeclient.api.v1.watch.namespace(namespace).configmaps.getObjectStream();
   configmapStream.on('data', event => {
+    console.log(event);
     if (event.object && event.object.metadata.name.endsWith(DATASYNC_CONFIGMAP_SUFFIX)) {
       updateAll(namespace, kubeclient);
+    } else if (event.type === 'DELETED') {
+      appNames = _.remove(appNames, event.object.metadata.name);
+      lookat = [];
     }
   });
 
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -134,13 +134,18 @@ async function updateAppsAndWatch(namespace, kubeclient) {
  * @param {*} kubeclient - kubernetes client
  */
 async function watchMobileClients(namespace, kubeclient) {
+  logAction('We are in side the Mobile client watch');
   const appStream = await kubeclient.apis[mobileClientCRD.spec.group].v1alpha1.watch
     .namespaces(namespace)
     .mobileclients.getObjectStream();
 
   appStream.on('data', event => {
-    if (event.type === 'ADDED') {
+    if (event.type === 'ADDED' && !_.includes(appNames, event.object.metadata.name)) {
+      console.log(event); // Trying to work out whats been in the event stream
       updateAll(namespace, kubeclient);
+    } else if (event.type === 'DELETED') {
+      appNames = _.remove(appNames, event.object.metadata.name);
+      lookat = [];
     }
   });
 
--- a/server/appServices.js
+++ b/server/appServices.js
@@ -82,7 +82,6 @@ function updateApp(namespace, app, kubeclient) {
     .then(result => {
       const shouldUpdate = result[0];
       logAction('At if for the should update');
-      console.log(result);
       if (shouldUpdate) {
         const newServices = result[1];
         app.status = {
@@ -90,12 +89,6 @@ function updateApp(namespace, app, kubeclient) {
           namespace,
           services: newServices
         };
-        logAction('Were are stuck here');
-        kubeclient.apis[mobileClientCRD.spec.group].v1alpha1
-          .namespace(namespace)
-          .mobileclients(appName)
-          .get()
-          .then(resp => console.log(resp));
         console.log(`Update services for app ${appName}`);
         return kubeclient.apis[mobileClientCRD.spec.group].v1alpha1
           .namespace(namespace)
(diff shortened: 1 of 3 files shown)


server/appServices.js
```diff
@@ -4,6 +4,7 @@ const androidVariantCRD = require('./android-variant-crd.json');
 const iosVariantCRD = require('./ios-variant-crd.json');
 const mobileSecurityServiceCRD = require('./mobile-security-crd.json');
 const { debounce } = require('lodash');
+const _ = require('lodash');
```
craicoverflow: lodash is already imported above this.

```suggestion
const { debounce, includes } = require('lodash');
```

server/appServices.js
```diff
@@ -31,14 +34,31 @@ const retryWatchDebounce = debounce((namespace, kubeclient) => {
...
   return kubeclient.apis[mobileClientCRD.spec.group].v1alpha1
     .namespace(namespace)
-    .mobileclients.get()
+    .mobileclient.get() // This call returns once for every app that has been created
     .then(resp => resp.body)
-    .then(mobileclientList => mobileclientList.items)
+    .then(mobileclientList => {
+      if (updating) {
+        mobileclientList.items.forEach(element => {
+          if (!_.includes(appNames, element.metadata.name)) {
```
craicoverflow: https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_includes

server/appServices.js
Boomatang: The reason for the importing lodash as `_` is as follows. When `includes` was directly import there was strange things happening. I was getting odd results. As I was exploring issues I felt that the importing the namespace for lodash was safer. I will keep this in mind for the final PR

server/appServices.js
```diff
@@ -92,13 +115,15 @@ function getServicesForApp(namespace, app, kubeclient) {
 
 async function updateAppsAndWatch(namespace, kubeclient) {
   connectionAttempts++;
-  updateAll(namespace, kubeclient).then(() => {
+  logAction('About to call updateAll()');
+  updateAll(namespace, kubeclient).then(pickles => {
```
ziccardi: Why that `pickles` parameter?

server/appServices.js
```diff
@@ -240,9 +266,8 @@ async function watchIosVariants(namespace, kubeclient) {
     .iosvariants.getObjectStream();
 
   iosVariantStream.on('data', event => {
-    if (event.object && event.object.status && !updating) {
-      updateAll(namespace, kubeclient);
-    }
+    logAction('Data stream from IoS');
```
ziccardi: this should be 'Data stream from **iOS**'

server/appServices.js
```diff
@@ -282,6 +306,64 @@ async function watchMobileSecurityApps(namespace, kubeclient) {
   });
 }
 
+function handleUPSEventData(event, suffix, namespace, kubeclient) {
+  if (event.type === 'ADDED' && event.object && event.object.metadata.name.endsWith(suffix) && !updating) {
+    const data = getApp(appCollect, getAppName(event));
+
+    const currentServices = data.status.services.filter(ser => ser.name === 'push');
+    const service = currentServices[0];
```
ziccardi: `const currentServices = data.status.services.filter(ser => ser.name === 'push');` can be removed:

```suggestion
    const service = data.status.services.find(ser => ser.name === 'push');
```

server/appServices.js
```diff
@@ -282,6 +306,64 @@ async function watchMobileSecurityApps(namespace, kubeclient) {
...
+
+    if (event.object.kind === 'AndroidVariant') {
+      if (service && service.config && typeof service.config.android !== 'undefined') {
```
ziccardi: why 
> if (service && service.config && **typeof** service.config.android **!== undefined**) {

instead of 

> if (service && service.config && service.config.android) {

 ?

server/appServices.js
```diff
@@ -282,6 +306,64 @@ async function watchMobileSecurityApps(namespace, kubeclient) {
...
+        console.log('AndroidVariant existing, no action required');
+      } else {
+        logAction('Adding a AndroidVariant to apps');
+        fullAppRefresh(namespace, kubeclient);
+      }
+    } else if (event.object.kind === 'IOSVariant') {
+      if (service && service.config && typeof service.config.ios !== 'undefined') {
```
ziccardi: See previous comment

ziccardi:
@Boomatang Added some comment. Going to test now.

server/appServices.js
Boomatang: It was my random naming when trying to figure out what was happening. I will refactor

server/appServices.js
Boomatang: Good question, I guess it don't need to be that way.

server/appServices.js
Boomatang: Right now it could be removed. But may cause is we are planning to allow users have their customs services in MDC. Which if my thinking is right we can't guarantee that the `push` service would be the first in the list hence the reason to filter.

ziccardi: Not sure I get you. With 

`data.status.services.find(ser => ser.name === 'push');`

and with the previous

`data.status.services.filter(ser => ser.name === 'push');`

We are filtering to get **only** push service.

Boomatang: Sorry I miss read suggestion, I will change `.filter` to be `.find`

ziccardi:
overseeded by https://github.com/aerogear/mobile-developer-console/pull/425

PR #416, opened 2019-09-09
Restart the server on kubectl init errors

ziccardi:
## Motivation

https://issues.jboss.org/browse/INTLY-3335

## What
Added a process.exit to the server on kubectl client initialization errors.

## Why
It happens sometimes after a fresh installation that CRDs are not ready yet when MDC starts, making the server fail the kubectl client initializations: restarting the pod fixes the issue.

Since pods are automatically restarted, exiting will solve the issue.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -119,6 +119,9 @@ async function initKubeClient() {
     return kubeclient;
   } catch (e) {
     console.error('Failed to init kube client', e);
+
+    // Exiting the server. Waiting 5 seconds to avoid strict loops
+    setTimeout(() => process.exit(1), 5000);
     return null;
   }
 }


server/server.js
```diff
@@ -119,6 +119,9 @@ async function initKubeClient() {
     return kubeclient;
   } catch (e) {
     console.error('Failed to init kube client', e);
+
+    // Exiting the server. Waiting 5 seconds to avoid strict loops
+    setTimeout(() => process.exit(1), 5000);
```
craicoverflow: I don't think this will solve the issue.

```sh
npm info it worked if it ends with ok
npm info using npm@6.9.0
npm info using node@v10.16.3
npm info lifecycle ui@0.1.0~prestart: ui@0.1.0
npm info lifecycle ui@0.1.0~start: ui@0.1.0

> ui@0.1.0 start /opt/app-root/src
> node server/server.js

Failed to init kube client Error: Failed to get /openapi/v2 and /swagger.json: Cannot convert undefined or null to object
    at _getSpec.catch.then.catch.err (/opt/app-root/src/node_modules/kubernetes-client/lib/swagger-client.js:58:15)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Check services for all apps Mon Sep 09 2019 10:54:16 GMT+0000 (Coordinated Universal Time)
Listening on port 4000
(node:16) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'apis' of null
    at updateAll (/opt/app-root/src/server/appServices.js:37:21)
    at updateAppsAndWatch (/opt/app-root/src/server/appServices.js:95:3)
    at run (/opt/app-root/src/server/server.js:148:3)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:16) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:16) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

As you can see from the logs, the root of the error occurs at server.js:148, which is outside of this try/catch block.

ziccardi: The `Failed to init kube client` error message is inside the `initKubeClient` function, that is called by the `run` function. In case of error `initKubeClient` was returning `null` thus explaining the `Cannot read property 'apis' of null` error

wei-lee:
@ziccardi you can try reproduce it by deleting one of the custom resources from the cluster perhaps?

PR #420, opened 2019-09-12
Aerogear 9890

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9890

## What
Add logging to the server unit of the MDC

## Why
To allow clear logging messages for the users

## How
By adding the winston js module to the package.json file

## Verification Steps
1. run `npm install` to get the latest packages
2. run the MDC both a production and non-production set up
3. view the logs and they show be a cleaner format
	`<timestamp(ISO 8601)>: <module> - <log level>: <Log message>`
4. in debug mode log file should be created locally
5. run `npm run lint`. There should be no warnings or errors

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/server/logger.js
+++ b/server/logger.js
@@ -1,0 +1,52 @@
+/*
+ ** When importing the logger the module name should be add to the logger
+ ** Example import:
+ ** const logger = require('./logger')('server.js');
+ ** 'server.js' been the file where the logger is been imported into.
+ */
+
+const { createLogger, format, transports } = require('winston');
+
+const messageFormat = format.printf(
+  ({ level, label, message, timestamp }) => `${timestamp}: ${label} - ${level}: ${message}`
+);
+
+module.exports = function(fileName) {
+  const defaultConsoleFormat = format.combine(
+    format.colorize({ all: true }),
+    format.timestamp(),
+    format.label({ label: fileName }),
+    messageFormat
+  );
+
+  const options = {
+    info: {
+      console: {
+        level: 'info',
+        format: defaultConsoleFormat
+      }
+    },
+    debug: {
+      console: {
+        level: 'debug',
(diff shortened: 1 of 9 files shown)


ziccardi (changes requested):

server/logger.js
```diff
@@ -0,0 +1,57 @@
...
+        level: 'debug',
+        handleExceptions: true,
+        format: format.combine(format.timestamp(), format.label({ label: fileName }), messageFormat)
+      }
+    }
+  };
+
+  const logger = createLogger();
+
+  if (process.env.NODE_ENV !== 'production') {
```
ziccardi: We will log to the console most of the time, so I would configure the `file transport` as optional
```suggestion
 if (process.env.NODE_ENV === 'development') {
    logger.add(new transports.Console(options.debug.console));
    logger.add(new transports.File(options.debug.file)); // creates local dev-debug.log file
  } else {
    logger.add(new transports.Console(options.info.console));
  }
```

Why do we have a `options.debug.console` and a `options.info.console` ? They looks identical

ziccardi (changes requested):

server/logger.js
Boomatang: Ok I will change this. I just want to point out that the `logger.verbose` calls will only show in the `debug` config. If required at a later stage we could add an extra flag for this.

Boomatang: I missed your question about the `options.debug.console` and options.info.console`. The difference is the log level which is shown in the console. `options.debug.console` will log all levels from `debug` up while `options.info.console` only logs from `info` up which means `debug` and `verbose` are not logged to the console.

Boomatang:
Closing

PR #425, opened 2019-09-12
fix memory issue when there are many apps

wei-lee:
## Motivation

Fix issue https://issues.jboss.org/browse/INTLY-3426 and https://issues.jboss.org/browse/AGMS-711. 

The root cause of the problem is that in the current implementation, there will be X number of concurrent calls to `updateAll` function, where `X` is equal to the number of apps. This is why we see that as the number of apps grow, the app will use more memory. The problem is that in `updateAll` function, it doesn't check if the `updating` flag is set to true.  While we could just add that check to fix the issue, there is a corner case won't be covered by this fix: if there is a changed happened while there is `updateAll` is being executed, this update change will be lost and it means there could be a delay for the mdc to show updated content.

So what we want is to make sure at any given time, there should be only one `updateAll` is being executed. Whenever there are updates received from the watchers, we also want to make sure they will invoke another `updateAll` call.

So the fix here is to introduce an internal queue. There will be only one worker to process the queue and the concurrency is set to one. Whenever there are changes from the watchers, a task will be pushed to the queue to ensure `updateAll` will be called at some point.

This change is deployed to here: https://console.rhmds-qe1.openshift.com/console/project/openshift-mobile-developer-console/browse/pods/mdc-9-zg7wd?tab=metrics.

@ziccardi can you review and verify this PR tomorrow morning. I have double the amount of apps in that cluster and you can check the memory usage history on this cluster to see the trend. If you are happy with it, feel free to merge and cut another release. Also let's update the operator to increase to memory limit to 256MB for MDC just in case.

Thanks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1622,8 +1622,7 @@
     "@types/d3-color": {
       "version": "1.2.2",
       "resolved": "https://registry.npmjs.org/@types/d3-color/-/d3-color-1.2.2.tgz",
-      "integrity": "sha512-6pBxzJ8ZP3dYEQ4YjQ+NVbQaOflfgXq/JbDiS99oLobM2o72uAST4q6yPxHv6FOTCRC/n35ktuo8pvw/S4M7sw==",
-      "optional": true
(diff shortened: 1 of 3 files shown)


ziccardi:
👍 Great!
ping @Boomatang

PR #436, opened 2019-10-01
Aerogear 9987

Boomatang:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-9987

## What
Add support in the MDC for web push with UPS

## Why
Web push is a new feature in UPS

## How
By following the similar steps of the other push features 

## Verification Steps
1. Have MDC running with UPS that supports web push
2. Create an app 
3. Bind UPS using web push.
4. Check that the `mobile-service.json` gets updated
5. Check icon is shown on the home page.
6. Create new app
7. Bind both Android and iOS.
8. Check the option to bind web push is available
9. Bind the web push
10. All 3 binds should have been done.
11. Check, The `Bind to App` button should be gone at this stage.
12. Check the `mobile-service.json` should list the config details for all 3 bindings
13. Check that the web push binding can be removed from the app.
14. Run `npm run test`, all test should pass

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/server/mobile-services.js
+++ b/server/mobile-services.js
@@ -142,7 +142,7 @@ const PushService = {
             }
 
             // get the AndroidVariant and add it
-            const androidVariant = variants.find(v => !!v && v.kind === ANDROID_VARIANT_KIND);
+            const androidVariant = variants.find(v => v && v.kind === ANDROID_VARIANT_KIND);
             if (androidVariant && androidVariant.status) {
               push.config[ANDROID_VARIANT_TYPE] = {
                 variantSecret: androidVariant.status.secret,
@@ -151,7 +151,7 @@ const PushService = {
             }
 
             // get the IOSVariant and add it
-            const iosVariant = variants.find(v => !!v && v.kind === IOS_VARIANT_KIND);
+            const iosVariant = variants.find(v => v && v.kind === IOS_VARIANT_KIND);
             if (iosVariant && iosVariant.status) {
               push.config[IOS_VARIANT_TYPE] = {
                 variantSecret: iosVariant.status.secret,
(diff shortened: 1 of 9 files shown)


server/mobile-services.js
```diff
@@ -134,6 +159,15 @@ const PushService = {
               };
             }
 
+            // get the AndroidVariant and add it
+            const webPushVariant = variants.find(v => !!v && v.kind === WEB_PUSH_VARIANT_KIND);
```
ziccardi: The `!!v &&` is there in the `!!v && v.kind` just to ensure that `v` is not null. Since we are not returning the value, there is no advantage in making the double negation. This can simply be `v && v.kind

```suggestion
            const webPushVariant = variants.find(v => v && v.kind === WEB_PUSH_VARIANT_KIND);
```

src/models/mobileservices/pushvariantcr.js
```diff
@@ -229,6 +272,45 @@ export class PushVariantCR extends CustomResource {
...
+            platformConfig: {
+              alias: {
+                validation_rules: [
+                  {
+                    type: 'required',
+                    error: 'Web push requires an Alias'
+                  }
+                ]
+              },
+              publicKey: {
```
ziccardi: I believe we could add a validator to validate that the passed in object are really a public and a private key

Boomatang: This we can do. How do we validate the public and private keys, as in what format should we be checking for.

server/mobile-services.js
Boomatang: The check `!!v` is also done on both Android and iOS variants. Should I remove this from them.

ziccardi:
Taken over in #437

PR #437, opened 2019-10-01
Aerogear 9987

ziccardi:
## [WIP] Motivation
https://issues.jboss.org/browse/AEROGEAR-9987

## What
Add support in the MDC for web push with UPS

### Changes:

- Added a new wizard page for the WebPush variant
- Added the following fields validator rules:
    - CompositeRule: it is a new rule that allow to specify a list of subrules and an evaluation algorithm whose value can be either `all` (all of the subrules must succeed) or `any` (at least one of the subrules must succeed)
   - ValidEmailRule
   - ValidUrlRule
   - ValidMailToRule
   - ValidECDSARule

## Why
Web push is a new feature in UPS

## How
By following the similar steps of the other push features 

## Verification Steps
1. Have MDC running with UPS that supports web push
2. Create an app 
3. Bind UPS using web push.
4. Check that the `mobile-service.json` gets updated
5. Check icon is shown on the home page.
6. Create new app
7. Bind both Android and iOS.
8. Check the option to bind web push is available
9. Bind the web push
10. All 3 binds should have been done.
11. Check, The `Bind to App` button should be gone at this stage.
12. Check the `mobile-service.json` should list the config details for all 3 bindings
13. Check that the web push binding can be removed from the app.
14. Run `npm run test`, all test should pass

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

### DIFF
--- a/server/mobile-services.js
+++ b/server/mobile-services.js
@@ -142,7 +142,7 @@ const PushService = {
             }
 
             // get the AndroidVariant and add it
-            const androidVariant = variants.find(v => !!v && v.kind === ANDROID_VARIANT_KIND);
+            const androidVariant = variants.find(v => v && v.kind === ANDROID_VARIANT_KIND);
             if (androidVariant && androidVariant.status) {
               push.config[ANDROID_VARIANT_TYPE] = {
                 variantSecret: androidVariant.status.secret,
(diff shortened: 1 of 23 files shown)


server/mobile-services.js
```diff
@@ -126,14 +151,23 @@ const PushService = {
...
               };
             }
 
+            // get the AndroidVariant and add it
+            const webPushVariant = variants.find(v => v && v.kind === WEB_PUSH_VARIANT_KIND);
+            if (webPushVariant && webPushVariant.status) {
+              push.config[WEB_PUSH_VARIANT_TYPE] = {
+                variantSecret: webPushVariant.status.secret,
+                variantId: webPushVariant.status.variantId
+              };
```
secondsun: appServerKey Needs to be added here as well.

wei-lee:
we should update binding screen to allow provide help information on the input fields so that users will have better understanding how to fill in the value for these fields. But this is not specific to web push, just something we should address in general.

secondsun:
@wei-lee should I make a JIRA ticket to update those fields in the MDC?  Since it isn't webpush specific I'd like to keep this PR from branching out from its "core".

PR #438, opened 2019-10-03
Add versions to about modal

jessiehuff:
Added versions to the about modal
(Also cleaned up a few unnecessary comments.)

Image 10-7-19 at 1 32 PM


dlabaj:
@jessiehuff Looks like there is some linting issues in this PR.

PR #441, opened 2019-10-09
Disable initial form validation until form changed

jessiehuff:
Fixed form validation so that there is a blank form on entry, validates after a user inputs data with live validation, and keeps the disabled button until criteria is met for form submission

Initial form:
Image 10-9-19 at 9 22 AM

After incorrect entry:
Screen Shot 2019-10-09 at 9 23 16 AM

After correct entry:
Image 10-9-19 at 9 40 AM

### DIFF
--- a/src/components/mobileservices/BindingPanel.js
+++ b/src/components/mobileservices/BindingPanel.js
@@ -29,7 +29,8 @@ export class BindingPanel extends Component {
       validationRules,
       isFormValid: false,
       onChangeHandler,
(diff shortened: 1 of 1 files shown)


christiemolloy:
Forms look awesome Jessie, when I change push server from "android" to "iOS" you'll see that the validation starts from the beginning. This could be captured in another issue because there is a problem with iOS and the dropdown... where it throws an error

Screen Shot 2019-10-23 at 10 16 38 AM

PR #444, opened 2019-10-17
Adding proxy

ziccardi:
## Motivation
https://issues.jboss.org/browse/AEROGEAR-10028


package.json
```diff
@@ -10,6 +10,7 @@
     "coveralls": "3.0.6",
     "elliptic": "^6.5.1",
     "express": "4.17.1",
+    "express-http-proxy": "^1.6.0",
```
wei-lee: it doesn't look like this module is used anywhere? Is it actually required?

package.json
```diff
@@ -58,15 +60,15 @@
...
-    "http-proxy-middleware": "0.19.1",
     "axios-mock-adapter": "1.17.0",
+    "babel-eslint": "10.0.3",
     "cors-anywhere": "0.4.1",
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.14.0",
     "eslint-plugin-patternfly-react": "0.3.0",
     "eslint-plugin-prettier": "3.1.0",
     "fetch-mock": "7.3.9",
+    "http-proxy-middleware": "^0.20.0",
```
wei-lee: I think this module should be move to `dependencies` section?

server/helpers.js
```diff
@@ -1,6 +1,6 @@
-function addProtocolIfMissing(url) {
-  if (url && !url.startsWith('http')) {
-    return `https://${url}`;
+function addProtocolIfMissing(url, protocol = 'https') {
+  if (url && !url.startsWith(protocol)) {
+    return `${protocol}://${url}`;
```
wei-lee: if the url starts with `http://`, but the protocal is `https`, this function will end up with adding the protocol again.

ziccardi:
@wei-lee Your comment should be addressed. It is deployed at https://mdc-mdc-proxy-mobile-developer-console.apps.gotham-3f65.open.redhat.com

server/server.js
```diff
@@ -15,9 +14,36 @@ const androidVariantCRD = require('./android-variant-crd.json');
...
+  target: addProtocolIfMissing(process.env.OPENSHIFT_HOST, 'https'),
+  secure: false,
+  logLevel: 'warn',
+  changeOrigin: true,
+  ws: true,
+  pathRewrite: {
+    [`^${OPENSHIFT_PROXY_PATH}`]: '' // remove base path
+  },
+  onProxyReq: (proxyReq, req) => {
+    proxyReq.setHeader('authorization', `Bearer ${req.headers['x-forwarded-access-token']}`);
```
wei-lee: just thinking, does this work in dev environment? It might be no such header when there is no oauth-proxy running?

